PR #123, opened 2014-06-16
Type hinting generates better error messages when an invalid session is ...

Ashpants:
...supplied

Otherwise if a different object or null is supplied, the Facebook request constructor throws an undefined index on the session object.


soffieswan041:
> ...supplied
> 
> Otherwise if a different object or null is supplied, the Facebook request constructor throws an undefined index on the session object.

soffieswan041:
> Thank you for signing our Contributor License Agreement. We can now accept your code for this (and any) Facebook open source project. Thanks!

PR #314, opened 2014-12-31
Fix auth exceptions

yguedidi:
How we don't see this before???

All current ``FacebookAuthorizationException`` are in fact authentication exceptions!
I also rename ``FacebookPermissionException`` to ``FacebookAuthorizationException`` because security is often authentication and authorization.


SammyK:
:+1: :)

PR #316, opened 2015-01-06
Prevent generation of logout url with app session.

gfosco:
Using getLogoutUrl with an app session can leak the appSecret.

### DIFF
--- a/src/Facebook/Entities/AccessToken.php
+++ b/src/Facebook/Entities/AccessToken.php
@@ -374,10 +374,7 @@ class AccessToken
    */
   public function isAppSession()
   {
-    if (strpos($this->accessToken, "|") !== FALSE) {
-      return true;
-    }
-    return false;
+    return strpos($this->accessToken, "|") !== false;
   }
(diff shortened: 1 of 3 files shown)


SammyK:
Wow - nice catch!

src/Facebook/Entities/AccessToken.php
```diff
@@ -367,4 +367,17 @@ public function __toString()
...
   }
 
+  /**
+   * Returns true if the access token is an app session token.
+   *
+   * @return bool
+   */
+  public function isAppSession()
+  {
+    if (strpos($this->accessToken, "|") !== FALSE) {
```
SammyK: Could be just one line with `false` in lowercase. :)

```php
return strpos($this->accessToken, '|') !== false;
```

gfosco:
Thanks for the review, going to merge and tag 4.0.15.

SammyK:
:+1:

yguedidi:
:+1:

PR #323, opened 2015-01-16
Add GetProfilePicture function

zoneroot:
This function returns the profile picture url of the user.

### DIFF
--- a/src/Facebook/GraphUser.php
+++ b/src/Facebook/GraphUser.php
@@ -132,4 +132,17 @@ class GraphUser extends GraphObject
     return $this->getProperty('location', GraphLocation::className());
   }
 
+  /*
+  *Returns the user profile picture url
+  *
+  *@return string
+  */
+  public function getProfilePicture($width = null, $height = null)
+  {
+    $url = 'http://graph.facebook.com/'. $this->getProperty('id') .'/picture';
+    if ($width & $height)
+      $url .= '?width='. $width .'&height='. $height;
+    return $url; 
+  }
+
 }


roman02:
Is it possible to pass only width **or** height to get a profile pic?

zoneroot:
No, if you want to set size you need to define width and height, but if you don't send params, the image size is set by the graph API.

SammyK:
Thanks for the contribution @zoneroot! There are a few issues with the way this is implemented.

The `picture` field is actually an [edge](https://developers.facebook.com/docs/graph-api/reference/v2.2/user/picture/) on the user node. So the picture should be represented as its own GraphNode object. In v4.1 of the SDK, the picture field on the user node will be [automatically cast](https://github.com/facebook/facebook-php-sdk-v4/blob/master/src/Facebook/GraphNodes/GraphUser.php#L42) as a [GraphPicture](https://github.com/facebook/facebook-php-sdk-v4/blob/master/src/Facebook/GraphNodes/GraphPicture.php) node. So the API for accessing a user's profile picture would be something like this:

```php
$pic = $userNode->getPicture();

if ($pic) {
  $picUrl = $pic->getUrl();
}
```

What you've submitted here would work a bit better as a helper of some sort but shouldn't belong on the GraphUser node.

Not sure what @gfosco wants to do with this since we're so close to 4.1 launch, but having a helper for a user's profile pic might be useful. I'd certainly recommend sending the picture request over https instead of http. Or even better using the constant `FacebookRequest::BASE_GRAPH_URL`. You'll also need to adjust the syntax a bit to be compliant with the rest of the SDK (spacing in the doc blocs & curly brackets in the `if` statement).

yguedidi:
@SammyK it's an edge if it's called with `?redirect=false`.

Without it it's a 302 redirection to the picture's URL. It might be useful to be able to get that "simple" URL just to display the picture instead of making a graph call to get the real picture URL.

Maybe a `->getPicture()` that returns the edge object, and another `->getPictureUrl()` that returns the simple URL? :)

SammyK:
@yguedidi It's still an edge if it's called with `?redirect=true` as well. It's just that the response is a little different. :)

> Maybe a ->getPicture() that returns the edge object

Yeah, that's how it's set up in 4.1 - it returns a `GraphPicture`.

> and another ->getPictureUrl() that returns the simple URL? :)

I'm all for adding a quick-n'-easy way of generating a profile pic URL, but I'm not sure the `GraphUser` node is the best place for it. All the `GraphObject`'s (soon to be `GraphNodes` hopefully!) represent the Graph API nodes & edges precisely. If we start creating methods inside of those that don't actually exist in the Graph API we start to move way from a nice domain-driven design style.

I would recommend making a UserHelper or something with a `getPictureUrl()` method that accepts a `GraphUser` node or `id` as an argument and generates the URL from that.

It's hard to figure out where stuff should go! :)

yguedidi:
> It's just that the response is a little different. :)

:+1:

But as you said, in both cases it's an edge, so, basicly, should be accessible through `$userNode->getPicture()`.

What do you think about something like:
```php
public function getPicture($edge = false)
{
    if (!$edge) {
        return 'https://graph.facebook.com/'. $this->getProperty('id') .'/picture';
    }

    return $this->getProperty('picture');
}
```

SammyK:
Nice idea! That would still represent the Graph API model since it also returns the picture URL in the form of a string (`Location:` header redirect) if you pass `redirect=true`.

I'd tweak it a bit:

```php
public function getPicture($asEdge = false, $params = [], $separator = '&')
{
    if ( ! $edge) {
        $query = $params ? '?' . http_build_query($params, null, $separator) : '';

        return FacebookRequest::BASE_GRAPH_URL . '/' . $this->getProperty('id') . '/picture' . $query;
    }

    return $this->getProperty('picture');
}
```

What do you think of that? :)

yguedidi:
I'd tweak it too:
```php
public function getPicture($asEdge = false, array $params = [], $asHTML = false)
{
  if (!$asEdge) {
    $query = $params ? '?' . http_build_query($params, null, $asHTML ? '&amp;' : '&') : '';

    return FacebookRequest::BASE_GRAPH_URL . '/' . $this->getProperty('id') . '/picture' . $query;
  }

  return $this->getProperty('picture');
}
```
- 2 spaces indentation
- array type hint
- less customizable separator

:stuck_out_tongue_closed_eyes:

But for real, there is a problem here, `$params` and `$asHTML` (or `$separator`) are used only when it's not an edge.

What if I call `$user->getPicture(false, ['foo' => 'bar'], true /* or '&amp;' */)`?
Should I receive the node with the `url` field using `$amp;` and containing `foo=bar`?

Back to 2 functions way...

yguedidi:
@SammyK thoughts on this? :)

SammyK:
Lol - indeed: 2-spaces indentation... :)

Yeah, I think we're butting up against two conflicting ideas:

1) Keep the API as simple and easy to use as possible.
2) Keep a strict DDD & SOLID approach.

Idea two makes things slightly more complicated for the developer using the SDK, because it requires they know more about the "Facebook way" of doing things. But the up side is that this will help people better understand Facebook's domain and best practices.

As a general rule, I'm thinking we adhere to no. 2 as much as possible and then make a helper or service to address idea #1. Which I feel we've done a pretty good job at so far with the `Facebook\Facebook` super service.

In this particular case, I think you're right. We should go back to the two-method approach as that would be more "correct". And maybe we could provide some sort of helper to bridge the gap between no. 1 & no. 2 to make it super easy for people to grab a profile picture for a user.

What do you think? :)

yguedidi:
I think the question that we have to ask ourselves is: What is the domain?

- In one case we have a string, an URL that point to an image.
- In the other case we have a JSON object that give data about that image.

Two different things, two domain object, two functions! And no conception problems :)

SammyK:
I agree - a node should be returned as a `GraphNode` and a string should be returned as a plain old PHP `string`.

But I think I can see what you're saying now about the difference between `/user/picture` and `/user/picture?redirect=true`. In both cases `/picture` is still an edge according to the Graph API but in the `?redirect=true` case it doesn't behave like an edge. So it really should be called something else IMO. :)

Usually a one-to-one relationship between two nodes in the Graph API is accessed from a field like the [`significant_other` field on the user node](https://developers.facebook.com/docs/graph-api/reference/v2.2/user) which returns another user node. But it seems the Graph API people had to cheat a bit on the user-to-picture node and added a "connivence edge" so that people could easily display user & [page](https://developers.facebook.com/docs/graph-api/reference/v2.2/page/picture/) pictures on their site without having to parse JSON.

Since the user and page nodes seem to be the only two nodes that have a "connivence edge" (if you know of another one, let me know!) maybe the SDK could provide a "connivence method" to access these special cases.

So we've come full circle! It seems like the best solution would be:

```php
// Plain old picture node
$pictureNode = $userNode->getPicture();

// "Connivence method" to support the "connivence edge"
$pictureUrl = $userNode->getPictureUrl($params, $separator);
```

Phew! Thoughts? :)

SammyK:
@zoneroot Did you ever think your original PR would cause so much discussion? Lol.

yguedidi:
`/picture` is an edge, a relationship between a user/page/event/album/app and a file. And with `redirect=false`, it's something like "I want to get metadata about the picture file" :)

SammyK:
Yeah, as I was thinking about it later, the one-to-only-one relationships that you can access through a node's fields like the `hometown` field on the User node (which returns a Page node) for example, are actually the "cheats". You should technically have to access those nodes from an edge to strictly conform to Graph Theory, but it's a lot more convenient to retrieve them as fields.

gfosco:
I'll close this since we're so close on 4.1... Do we have this handled in the new version?

SammyK:
We still need to address this but it'd be good to open a new one after 4.1 is launched. :)

PR #325, opened 2015-01-21
allow to force re-authenticate with getLoginUrl method

vrcAlbert:
I'm not sure if it's the proper way to do it, but I need to force authentication for users using my app. This does the job, based on https://developers.facebook.com/docs/facebook-login/reauthentication?locale=fr_FR

Thanks in advance for considering it.

### DIFF
--- a/src/Facebook/FacebookRedirectLoginHelper.php
+++ b/src/Facebook/FacebookRedirectLoginHelper.php
@@ -88,7 +88,7 @@ class FacebookRedirectLoginHelper
    *
    * @return string
    */
-  public function getLoginUrl($scope = array(), $version = null, $displayAsPopup = false)
+  public function getLoginUrl($scope = array(), $version = null, $displayAsPopup = false, $reauthenticate = false)
   {
     $version = ($version ?: FacebookRequest::GRAPH_API_VERSION);
     $this->state = $this->random(16);
@@ -100,6 +100,10 @@ class FacebookRedirectLoginHelper
       'sdk' => 'php-sdk-' . FacebookRequest::VERSION,
       'scope' => implode(',', $scope)
     );
+
+    if (!empty($reauthenticate)) {
+      $params['auth_type'] = $reauthenticate === true ? 'reauthenticate' : $reauthenticate;
+    }
     
     if ($displayAsPopup)
     {


gfosco:
There is an existing method, getReRequestUrl..  Does this not achieve what you're looking for?

vrcAlbert:
Hello,

Thanks for considering my PR!

It's not about asking again for missing permissions; I need to reauthenticate users, i.e. having the login form displayed even if a user is currently logged into facebook.

Here's what I found about this feature : https://developers.facebook.com/docs/facebook-login/reauthentication

src/Facebook/FacebookRedirectLoginHelper.php
```diff
@@ -100,6 +100,10 @@ public function getLoginUrl($scope = array(), $version = null, $displayAsPopup =
       'sdk' => 'php-sdk-' . FacebookRequest::VERSION,
       'scope' => implode(',', $scope)
     );
+
+    if (!empty($reauthenticate)) {
```
gfosco: Why the empty() check rather than just truthiness?  And once inside the if, why the ternary condition instead of just setting it.

vrcAlbert: Considering documentation, "auth_type" parameter only accept two values which are "reauthenticate" and "https". Rather than passing "null" or "false" and having a result I can't predict, I added the first "if" condition.

The second one allow to use "true" to set "reauthenticate", or passing one of the two expected values ("reauthenticate" or "https").

gfosco: Ok, good answers.  Thanks :)

gfosco:
Take a look at the master branch and consider adding a PR for the new version.  It is nearing release.  Thanks for your contribution!

SammyK:
The master branch handles this with three methods on the `FacebookRedirectLoginHelper`:

```php
$url = $helper->getLoginUrl(/* . . . */);
$url = $helper->getReRequestUrl(/* . . . */);
$url = $helper->getReAuthenticationUrl(/* . . . */);
```

So we're good. :)

gfosco:
:+1:

vrcAlbert:
Thanks!

Issue #320, opened 2015-01-13
[6.0] [Proposal] Rename GraphObject & GraphList


SammyK:
After discussing the Facebook PHP SDK a bit on the [domain-driven design discussion on the PHP Roundtable](https://www.phproundtable.com/episode/domain-driven-design-in-php), I realized that there are a few places in the SDK where we could better be in sync with Facebook nomenclature.

The two biggest DDD-eye-sores are the naming `GraphObject` & `GraphList`.

1. The `GraphObject` represents a "thing" on the Graph API. ["Things" in Graph are called "nodes"](https://developers.facebook.com/docs/graph-api/quickstart/v2.2#basics). We started going the right direction by renaming the containing folder to `GraphNodes` but let's go ahead and complete that by renaming `GraphObject` to `GraphNode`.
2. Relationships between nodes in Graph [are called "edges"](https://developers.facebook.com/docs/graph-api/quickstart/v2.2#basics).The `GraphList` represents one or more nodes returned by Graph from an edge. I recommend we rename `GraphList` to `GraphEdge`.

This change will disambiguate what these objects represent and reduce confusion when someone is reading through the Graph API and Facebook SDK documentation. :)

yguedidi:
:+1:

yguedidi:
@SammyK as we discuted in #323, edges, like the picture one, are not always lists...

SammyK:
Yep! One more reason to rename it to `GraphEdge`. :)

SammyK:
Another way of thinking about this is that edges describe relationships between nodes. In some cases that relationship will always be between a node and one-and-only-one other node.

@gfosco In an effort to make the SDK more DDD-focused, do you know (or someone from the Graph API knows) if there is a name to describe an edge that will always return one-and-only one node? Like in the `user/picture` example?

gfosco:
They are just called edges, doesn't appear to be any specific language around an edge that only returns one node.  I support the renaming.

SammyK:
Sounds good! Changed the title to slate this for 4.2 :)

yguedidi:
I think there is a simple solution:
- keep `{id}/picture` that will always redirect to the picture. Nor a node, nor an edge, just a shortcut to the picture when you got the id.
- add a `picture` field to the node that will represent the actual object we get with `?redirect=false`

Thoughts?

SammyK:
Yeah, that sounds good!

argb:
support renaming

[...]

PR #330, opened 2015-01-27
[4.1] Refactored OAuth 2.0 functionality

SammyK:
Finally got around the access token refactor! :)

So I finally figured out what has been bothering me about the access token handling.

- There was too much focus on specifically access tokens & not the whole OAuth 2.0 authentication process.
- This coupled the OAuth HTTP calls to the `AccessToken` entity violating the **S** in [SOLID](http://en.wikipedia.org/wiki/SOLID_%28object-oriented_design%29).
- This also coupled part of the OAuth process to the `FacebookRedirectLoginHelper`.
- It also incorrectly assumed access token metadata was a Graph node.

Also, the "get code from token" functionality was broken.

This PR addresses all these issues.

## Summary

- Moved all the access token stuff into the `Authentication` folder. There's more authentication-related stuff that we should add to the SDK later on (like better permissions checking/handling, etc.).
- Refactored `GraphSessionInfo` into `AccessTokenMetadata` since [the `/debug_token` endpoint](https://developers.facebook.com/docs/graph-api/reference/v2.2/debug_token) does not return a Graph node, but metadata from an access token.
	- Updated methods on `AccessTokenMetadata` to reflect all the metadata returned from the `/debug_token` endpoint.
- Refactored & removed `AccessToken::validateAccessToken()`
	- Added `AccessToken::validateMachineId('{machine-id}')` which throws a `FacebookSDKException`.
	- Added `AccessTokenMetadata::validateAppId('{app-id}')`, `AccessTokenMetadata::validateUserId('{user-id}')`, and `AccessTokenMetadata::validateExpiration()` as recommended [by the docs](https://developers.facebook.com/docs/facebook-login/manually-build-a-login-flow#checktoken). Those throw a `FacebookSDKException`.
- Added an `OAuth2Client` service which handles the OAuth 2.0 process. (Moved all the request-making methods from `AccessToken` to this class & named them based on Facebook's domain. Also killed the static methods.)
- Added `asString()` convenience method to `AccessToken`
- `OAuth2Client` is now constructor-injected into all of the helpers
- Added an `isAppAccessToken()` method to `AccessToken`
- Ported over #316 security fix from 4.0
- Added `getOAuth2Client()` convenience method to the `Facebook` super service
- Hot fixed a failing test in the `FacebookRedirectLoginHelperTest`.

## Examples

```php
/*
	Obtaining access tokens from helpers is easier (no need to get & inject FacebookClient)
*/
// OR: getJavaScriptHelper() or getCanvasHelper() or getPageTabHelper()
$helper = $fb->getRedirectLoginHelper();
$accessToken = $helper->getAccessToken();

echo (string) $accessToken;
// Or
echo $accessToken->asString();

/*
	Access Token handling is prettier
*/
// OAuth 2.0 client handler
$oAuth2Client = $fb->getOAuth2Client();

// Returns a link for "log in with Facebook"
$logInUrl = $oAuth2Client->getAuthorizationUrl('{redirect-url}', ['email'], 'foo_state');

// Returns an access token from a code
$accessToken = $oAuth2Client->getAccessTokenFromCode('{code}');

// Exchanges a short-lived access token for a long-lived one
$longLivedAccessToken = $oAuth2Client->getLongLivedAccessToken('{access-token}');

// Returns a code from a long-lived access token
$code = $oAuth2Client->getCodeFromLongLivedAccessToken('{access-token}');

try {
	$accessToken->validateMachineId('{machine-id}');
} catch (FacebookSDKException $e) {
	echo 'Invalid machine ID';
	exit;
}

/*
	Debugging the access token is more explicit (per docs)
*/
//Calls the `/debug_token` endpoint and returns an `AccessTokenMetadata` entity
$tokenMetadata = $oAuth2Client->debugToken('{access-token}');

try {
	$tokenMetadata->validateAppId('{app-id}');
} catch (FacebookSDKException $e) {
	echo 'Unexpected app ID in access token metadata';
	exit;
}

try {
	$tokenMetadata->validateUserId('{user-id}');
} catch (FacebookSDKException $e) {
	echo 'Unexpected user ID in access token metadata';
	exit;
}

try {
	$tokenMetadata->validateExpiration();
} catch (FacebookSDKException $e) {
	echo 'Access token has expired';
	exit;
}
```

Wudya think? :)

After this PR & feedback from @gfosco on #320, it should be just a matter of updating the docs and releasing 4.1!!


src/Facebook/Authentication/AccessToken.php
```diff
@@ -0,0 +1,223 @@
...
+ */
+class AccessToken implements \Serializable
+{
+
+  /**
+   * The access token.
+   *
+   * @var string
+   */
+  protected $accessToken = '';
```
yguedidi: `$value` is maybe a better name

src/Facebook/Authentication/AccessToken.php
```diff
@@ -0,0 +1,223 @@
...
+
+  /**
+   * A unique ID to identify a client.
+   *
+   * @var string|null
+   */
+  protected $machineId;
```
yguedidi: I think the machine id feature should be removed from the SDK. It's only used from client platforms.

src/Facebook/Authentication/AccessToken.php
```diff
@@ -0,0 +1,223 @@
...
+  {
+    return hash_hmac('sha256', $this->accessToken, $appSecret);
+  }
+
+  /**
+   * Setter for expires_at.
+   *
+   * @param int $timeStamp
+   */
+  protected function setExpiresAtFromTimeStamp($timeStamp)
```
yguedidi: protected method should go to the bottom of the class

src/Facebook/Authentication/AccessToken.php
```diff
@@ -0,0 +1,223 @@
...
+  {
+    if ($this->isAppAccessToken()) {
+      return true;
+    }
+
+    if ($this->expiresAt) {
+      return $this->expiresAt->getTimestamp() > time() + (60 * 60 * 2);
+    }
+
+    return false;
```
yguedidi: I'd invert the 2 above returns

src/Facebook/Authentication/AccessToken.php
```diff
@@ -0,0 +1,223 @@
...
+  {
+    if ($this->isAppAccessToken()) {
+      return false;
+    }
+
+    if ($this->getExpiresAt() instanceof \DateTime) {
+      return $this->getExpiresAt()->getTimestamp() < time();
+    }
+
+    return null;
```
yguedidi: Invert here too

src/Facebook/Authentication/AccessToken.php
```diff
@@ -0,0 +1,223 @@
...
+ */
+namespace Facebook\Authentication;
+
+use Facebook\Exceptions\FacebookSDKException;
+
+/**
+ * Class AccessToken
+ * @package Facebook
+ */
+class AccessToken implements \Serializable
```
yguedidi: `Serializable` can be removed, as it's not necessary, there is no resource or something else in the class. (My mistake, sorry)

src/Facebook/Authentication/AccessToken.php
```diff
@@ -0,0 +1,223 @@
...
+      throw new FacebookSDKException('Access token contains unexpected machine ID.', 401);
+    }
+  }
+
+  /**
+   * Returns the access token as a string.
+   *
+   * @return string
+   */
+  public function asString()
```
yguedidi: Is this really useful?

src/Facebook/Authentication/AccessTokenMetadata.php
```diff
@@ -0,0 +1,374 @@
...
+  }
+
+  /**
+   * Converts a unix timestamp into a DateTime entity.
+   *
+   * @param int $timeStamp
+   *
+   * @return \DateTime
+   */
+  private function convertTimestampToDateTime($timeStamp)
```
yguedidi: `$timestamp`

src/Facebook/Authentication/OAuth2Client.php
```diff
@@ -0,0 +1,298 @@
...
+    $params += [
+      'client_id' => $this->app->getId(),
+      'state' => $state,
+      'response_type' => 'code',
+      'sdk' => 'php-sdk-' . Facebook::VERSION,
+      'redirect_uri' => $redirectUrl,
+      'scope' => implode(',', $scope)
+    ];
+
+    return static::BASE_AUTHORIZATION_URL . '/' . $this->graphVersion . '/dialog/oauth?' .
```
yguedidi: What if the graph version is null? empty? You should validate it

src/Facebook/Authentication/OAuth2Client.php
```diff
@@ -0,0 +1,298 @@
...
+  /**
+   * Get the metadata associated with the access token.
+   *
+   * @param AccessToken|string $accessToken The access token to debug.
+   *
+   * @return AccessTokenMetadata
+   */
+  public function debugToken($accessToken)
+  {
+    $params = ['input_token' => (string) $accessToken];
```
yguedidi: I'd not use "magic" casting in the SDK, it's a useful feature, but I prefere to rely on "pure" functions in internal code.

src/Facebook/Authentication/OAuth2Client.php
```diff
@@ -0,0 +1,298 @@
...
+    ];
+
+    $response = $this->sendRequestWithClientParams('/oauth/client_code', $params, $accessToken);
+    $data = $response->getDecodedBody();
+
+    if (isset($data['code'])) {
+      return $data['code'];
+    }
+
+    throw new FacebookSDKException('Code was not returned from Graph.', 401);
```
yguedidi: I'd invert here too. My philosophy is: problems and early returns first, function purpose last :)

src/Facebook/Facebook.php
```diff
@@ -327,7 +350,7 @@ public function getRedirectLoginHelper()
    */
   public function getJavaScriptHelper()
   {
-    return new FacebookJavaScriptHelper($this->app);
+    return new FacebookJavaScriptHelper($this->app, $this->getOAuth2Client());
```
yguedidi: Passing app is useless, it's already owned by the OAuth2 client

src/Facebook/Helpers/FacebookRedirectLoginHelper.php
```diff
@@ -211,13 +191,24 @@ public function getLoginUrl($redirectUrl,
    * @param string $separator The separator to use in http_build_query().
    *
    * @return string
+   *
+   * @throws FacebookSDKException
    */
   public function getLogoutUrl($accessToken, $next, $separator = '&')
```
yguedidi: Is this still exists? It's no more documented

yguedidi:
@SammyK I did a quick review, I hope more tomorrow. Excellent work, as always ;)

SammyK:
Thanks so much @yguedidi! I'll do some refactoring tomorrow. :)

src/Facebook/Authentication/AccessToken.php
SammyK: I think it's important to keep in since you can use it in the scenario were you have your app running on multiple clients (like web, native mobile, etc). And if the same access token is being used across multiple devices, it's in danger of getting flagged for spam. In that case you would store a `code` which you use to generate new long-lived access tokens and create a unique access token for each client you have. And that code-for-token feature uses the `machine_id` for added security. So I think we should keep this feature. :)

src/Facebook/Authentication/AccessToken.php
SammyK: I'm curious why you're suggesting we remove it. I think it's useful for keeping the expire date and machine ID metadata.

src/Facebook/Authentication/AccessToken.php
SammyK: Yes - as described above when exchanging a code for a token. :)

src/Facebook/Authentication/OAuth2Client.php
SammyK: I'm not sure I understand what you mean. The `debugToken()` method is part of the public API:

```php
$oAuth2Client = $fb->getOAuth2Client();
$tokenMetadata = $oAuth2Client->debugToken('{access-token}');
```

A lot of times `{access-token}` will be an `AccessToken` entity that was returned from a `getAccessToken()` method or something. So this allows for either a string or `AccessToken` entity to be passed as an argument without the user having to type-cast the access token before passing it in.

src/Facebook/Facebook.php
SammyK: All of these helpers extend from the `FacebookSignedRequestFromInputHelper` which needs an instance of `FacebookApp` to inject to the `SignedRequest` entity. I considered pulling the `FacebookApp` from the `OAuth2Client` within `FacebookSignedRequestFromInputHelper`, but it felt wrong since signed requests and OAuth flow are two separate concepts that shouldn't depend on each other.

src/Facebook/Helpers/FacebookRedirectLoginHelper.php
SammyK: Yeah, I think this functionality is broken on FB's end, but I'd like to submit a bug request and double check it's no longer supported before killing it in the SDK. :)

SammyK:
Tweaks from @yguedidi complete (sans the ones I tried to clarify). :)

src/Facebook/Authentication/OAuth2Client.php
yguedidi: You're right. I just don't like implicit calls to `__toString()` in internal code :)

SammyK: What about this?

```php
$accessToken = $accessToken instanceof AccessToken ? $accessToken->asString() : $accessToken;
```

src/Facebook/Authentication/AccessToken.php
yguedidi: Because it's not necessary ([see here](http://3v4l.org/Frf5i)).

SammyK: Ah! I see what you're saying. Changed! :)

src/Facebook/Authentication/AccessToken.php
yguedidi: I mean, PHP code is executed server-side, so the secret key is secure and you can directly ask for an access token.
In the case of client-side platforms, they need to ask for a code (so we have to keep the `getCodeFromAccessToken` feature), but then, it the responsability of the client to exchange that code for a token, using the machine id. Server that excute PHP it not a client.

src/Facebook/Authentication/AccessToken.php
yguedidi: As you renamed the attribute to `$value`, I think `getValue()` makes more sense. `asString()` sementicaly duplicate the native `__toString()`

src/Facebook/Authentication/OAuth2Client.php
yguedidi: :+1: more explicit

src/Facebook/Helpers/FacebookRedirectLoginHelper.php
yguedidi: :+1:

src/Facebook/Facebook.php
yguedidi: I don't like that.. I'll try to find a solution to that problem tonight after my work.

src/Facebook/Authentication/AccessToken.php
SammyK: I guess I see what you're saying. I suppose we can kill this.

src/Facebook/Facebook.php
SammyK: Sounds good - I'm open to ideas! :)

SammyK:
Other tweaks complete. Just waiting to hear @yguedidi's ideas on injecting `FacebookApp` into `FacebookSignedRequestFromInputHelper`. :)

src/Facebook/Facebook.php
yguedidi: Still at work, but here a quick solution :)

As the OAuth2 client is used as an internat instance, with helper methods, I think the app should be the first parameters of each helper method like `getAuthUrl()` or `getATFromCode()`.
In fact, with the helper OAuth client, we can get an auth url for any app, so it should be a parameter.

SammyK: Sorry, I don't quite follow what you're saying. :rabbit:

yguedidi: Something like moving the `FacebookApp` dependency from the constructor to object methods :)

SammyK: I still don't follow. Sorry! lol. Can you provide a quick-and-dirty example? :)

yguedidi:
```php
class OAuthClient
{
    public function __construct(/*FacebookApp $app, */FacebookClient $client, $graphVersion = null);
    public function getAuthorizationUrl(FacebookApp $app, $redirectUrl, array $scope = [], $state, array $params = [], $separator = '&');
    protected function requestAnAccessToken(FacebookApp $app, array $params);
}

class FacebookRedirectHelper {
     public function __construct(FacebookApp $app,OAuth2Client $oAuth2Client, ...);
}

class FacebookSignedRequestFromInputHelper {
     public function getAccessToken()
     {
         return $this->oAuth2Client->getAccessTokenFromCode($this->app, $code);
     }
}
```

Quick and dirty, sorry

SammyK: Ah, I see what you mean. But since the `OAuthClient` methods are part of the public API, that means it would be up to the developer to inject the `FacebookApp` instance with each method. Which is quite inconvenient. Plus, injecting a dependency into every method is a code-smell indicating it should be constructor-injected.

But the way I've got it set up in this PR, the `FacebookApp` instance being injected into the signed request helpers and the instance injected into the `OAuth2Client` could be different and all kinds of bad things could happen once those two out-of-sync dependancies make it into the helper. 

How about all the helpers get a `FacebookApp` and `FacebookClient` on instantiation. Then the helpers create an instance of `OAuth2Client` by passing along those two dependancies. That way they are never our of sync and the developer doesn't have to inject a `FacebookApp` with every method call.

```php
class FacebookRedirectHelper {
     public function __construct(FacebookApp $app, FacebookClient $client) {
            $this->oAuth2Client = new OAuth2Client($app, $client);
            // . . 
     }
}
```

Would that work for you? :)

yguedidi: OK for me, I'll don't say that it's perfect, but I don't have an other solution, so go for it! :)

SammyK: Sounds good! Maybe once we get more eyeballs on the code we'll get some nicer dependency handling going forward. One step at a time! :) I'll refactor and push in a bit.

SammyK:
Refactored based on @yguedidi feedback! Should be good to merge pending @gfosco review! :)

gfosco:
This needs a rebase due to other PRs that were merged.. Sorry :)

SammyK:
I'm on it! :)

SammyK:
Fixed conflicts! :)

SammyK:
OMG! That means all we need is doc updates and we can tag a stable release!



Going to try and get a lot of this done while I'm at [Sunshine PHP](http://2015.sunshinephp.com/). :)

yguedidi:
@SammyK Congrats

SammyK:
Thanks @yguedidi! :)

PR #337, opened 2015-01-28
Adding GraphAchievement Node class

stevenmaguire:
In an effort to add approachable classes and objects to the library, and prompted by a laundry list item in https://github.com/SammyK/facebook-php-sdk-v4/issues/1, I have added a node class for "achievement" 

https://developers.facebook.com/docs/graph-api/reference/v2.2/achievement

This is my first contribution adding Node classes like this, I would like to know if there are any specific aspects of this implementation that should be refined before moving forward with adding additional classes.

I also created bare-bones class for the "application" node. This will need more energy later.

Finally, I created a base abstract `GraphNodeTest ` to be used to move some code repetition away from the individual Node test classes.

### DIFF
--- a/src/Facebook/GraphNodes/GraphAchievement.php
+++ b/src/Facebook/GraphNodes/GraphAchievement.php
@@ -51,7 +51,7 @@ class GraphAchievement extends GraphObject
     /**
      * Returns the user who achieved this.
      *
-     * @return string|null
+     * @return GraphUser|null
      */
     public function getFrom()
     {
@@ -61,7 +61,7 @@ class GraphAchievement extends GraphObject
     /**
      * Returns the time at which this was achieved.
      *
-     * @return string|null
+     * @return \DateTime|null
      */
     public function getPublishTime()
     {
@@ -71,7 +71,7 @@ class GraphAchievement extends GraphObject
     /**
      * Returns the app in which the user achieved this.
      *
-     * @return string|null
+     * @return GraphApplication|null
      */
     public function getApplication()
     {
@@ -82,7 +82,7 @@ class GraphAchievement extends GraphObject
      * Returns information about the achievement type this instance is
      * connected with.
      *
-     * @return string|null
+     * @return array|null
      */
     public function getData()
     {
@@ -105,10 +105,10 @@ class GraphAchievement extends GraphObject
      * Indicates whether gaining the achievement published a feed story for
      * the user.
      *
-     * @return boolean
+     * @return boolean|null
      */
     public function isNoFeedStory()
     {
-        return (bool) $this->getProperty('no_feed_story');
+        return $this->getProperty('no_feed_story');
     }
 }
--- a/src/Facebook/GraphNodes/GraphAchievement.php
+++ b/src/Facebook/GraphNodes/GraphAchievement.php
@@ -1,0 +1,114 @@
+<?php
+/**
+ * Copyright 2014 Facebook, Inc.
+ *
+ * You are hereby granted a non-exclusive, worldwide, royalty-free license to
+ * use, copy, modify, and distribute this software in source code or binary
+ * form for use in connection with the web services and APIs provided by
+ * Facebook.
+ *
+ * As with any software that integrates with the Facebook platform, your use
+ * of this software is subject to the Facebook Developer Principles and
+ * Policies [http://developers.facebook.com/policy/]. This copyright notice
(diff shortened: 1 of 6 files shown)


SammyK:
Thanks @stevenmaguire! You're busy with this laundry list! Are you gonna save anything for the open source workshop? Lol. :)

We might want to have a broader conversation of which nodes on the Graph API are ones we want to represent with a `GraphNode` subclass in the SDK. Cause there are a [heck-of-a lot of them](https://developers.facebook.com/docs/graph-api/reference) [citation needed].

And we might want to hold off on these types of PR's until after 4.1 (or 5.0?!) stable is released and #320 is addressed.

But this is good stuff!

src/Facebook/GraphNodes/GraphAchievement.php
```diff
@@ -0,0 +1,114 @@
...
+     */
+    public function getFrom()
+    {
+        return $this->getProperty('from');
+    }
+
+    /**
+     * Returns the time at which this was achieved.
+     *
+     * @return string|null
```
SammyK: Should be `\DateTime|null`

src/Facebook/GraphNodes/GraphAchievement.php
```diff
@@ -0,0 +1,114 @@
...
+     */
+    public function getId()
+    {
+        return $this->getProperty('id');
+    }
+
+    /**
+     * Returns the user who achieved this.
+     *
+     * @return string|null
```
SammyK: Should be `GraphUser|null`

src/Facebook/GraphNodes/GraphAchievement.php
```diff
@@ -0,0 +1,114 @@
...
+     */
+    public function getPublishTime()
+    {
+        return $this->getProperty('publish_time');
+    }
+
+    /**
+     * Returns the app in which the user achieved this.
+     *
+     * @return string|null
```
SammyK: Should be `GraphApplication|null`

src/Facebook/GraphNodes/GraphAchievement.php
```diff
@@ -0,0 +1,114 @@
...
+    public function getApplication()
+    {
+        return $this->getProperty('application');
+    }
+
+    /**
+     * Returns information about the achievement type this instance is
+     * connected with.
+     *
+     * @return string|null
```
SammyK: Should be `array|null`

src/Facebook/GraphNodes/GraphAchievement.php
```diff
@@ -0,0 +1,114 @@
...
+    public function getType()
+    {
+        return 'game.achievement';
+    }
+
+    /**
+     * Indicates whether gaining the achievement published a feed story for
+     * the user.
+     *
+     * @return boolean
```
SammyK: I'd change this to `boolean|null` and not cast the return value as `(bool)`. That way if it's `null` you would know that the field wasn't returned.

tests/GraphNodes/GraphNodeTest.php
```diff
@@ -0,0 +1,50 @@
...
+ */
+namespace Facebook\Tests\GraphNodes;
+
+use Mockery as m;
+use Facebook\GraphNodes\GraphObjectFactory;
+
+abstract class GraphNodeTest extends \PHPUnit_Framework_TestCase
+{
+    /**
+     * @var \Facebook\FacebookResponse
```
SammyK: Should be `\Facebook\FacebookResponse|\Mockery\MockInterface` so that the IDE's will know about mockery methods. :)

stevenmaguire:
Yeah, I'm not going to go through the whole list. I am just trying to find some things that give me good reason to learn more about the architecture. I am done for the moment :)

That is a good point on the number of `GraphNodes`. I noticed the same thing, and I wanted to get a conversation going to surface any other thoughts around how explicit we want to be with building them. It sounds like this conversation is looming!

SammyK:
Sweet! Glad to see you getting involved! :)

gfosco:
I'm going to accept this since @stevenmaguire went through the effort and it looks solid.  Thanks, we can argue over which nodes deserve subclasses later.

SammyK:
:+1: Excellent choice my friend. :)

stevenmaguire:
proud rooster

SammyK:
GraphAchievement unlocked! :unlock:

PR #341, opened 2015-02-03
Enhance #325

yguedidi:
- add missing php doc
- fix the case when you pass any value (eg `"foo"`)

@vrcAlbert sorry for being late to review your PR.. :/


gfosco:
Ok, I do like this better.  Thanks.

yguedidi:
@gfosco You're welcome

vrcAlbert:
@yguedidi No problem, somehow I know what it is :)
Thanks for the improvement!

PR #343, opened 2015-02-06
getTimezone for GraphUser

MauricioKruijer:
Since 'timezone' is considered a core element of the User object (and I need it) I've included it in GraphUser


MauricioKruijer:
signed.

gfosco:
Sure, I'll add this.  Not going to tag a release just yet, but it will eventually get in.  Thanks.

PR #348, opened 2015-02-10
Bringing Facbebook\PersistentData namespace into PSR2 compliance

stevenmaguire:
It's happening.

### DIFF
--- a/src/Facebook/PersistentData/FacebookMemoryPersistentDataHandler.php
+++ b/src/Facebook/PersistentData/FacebookMemoryPersistentDataHandler.php
@@ -29,28 +29,26 @@ namespace Facebook\PersistentData;
  */
 class FacebookMemoryPersistentDataHandler implements PersistentDataInterface
 {
+    /**
+     * @var array The session data to keep in memory.
+     */
+    protected $sessionData = [];
 
-  /**
-   * @var array The session data to keep in memory.
-   */
-  protected $sessionData = [];
+    /**
+     * @inheritdoc
+     */
+    public function get($key)
+    {
+        return isset($this->sessionData[$key])
+            ? $this->sessionData[$key]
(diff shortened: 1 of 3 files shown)


SammyK:
:+1: It's happening!

src/Facebook/PersistentData/FacebookMemoryPersistentDataHandler.php
```diff
@@ -29,28 +29,26 @@
...
-    $this->sessionData[$key] = $value;
-  }
+    /**
+     * @inheritdoc
+     */
+    public function get($key)
+    {
+        return isset($this->sessionData[$key])
+            ? $this->sessionData[$key]
+            : null;
```
yguedidi: One line IMO

stevenmaguire: It's a close call there. That line is 82 characters long on a single line. The linter does not object, so I refactored to one line.

screenshot

yguedidi: FYI, there is no hard limit in PSR-2 ;)

stevenmaguire: I know. I just like to keep lines short and legible.

yguedidi: For me, a ternary is more readable in one line, and it should always be in one line.
If it's too long, it mean you shouldn't use a ternary, but a classic `if`/`else`.
It's my personal tought

SammyK: I like wrapping my ternary's. :)

PR #355, opened 2015-02-11
Fix for invalid PHP syntax

SammyK:
This was causing the tests to fail. You can't concatenate string on a constant.

Now that we have Travis CI, we should make sure the tests are passing before merging in a PR. :)


yguedidi:
Included in #356 ;)

SammyK:
@gfosco Can you merge this one in as soon as possible? Everyone using the master branch is getting broken code. Yikes :exclamation:

yguedidi:
An example of bad split of an hardcoded string (ref [my comment](https://github.com/facebook/facebook-php-sdk-v4/pull/356#issuecomment-74104964) in #356) :satisfied:

SammyK:
Thanks @gfosco! Now my project builds can pass! :)

@yguedidi Troll troll troll!!!! :)

PR #357, opened 2015-02-13
clear the curl handle after closing it

mlebkowski:
The curl handle wasn’t cleared after closing, and due to this commit eb747316e313944a06b9129e51c58f5bfda2ef2d it wasn’t initialized again.

I also changed the `FacebookCurlHttpClient::$facebookCurl` field to be non-static (since every client has needs it’s own handle)

### DIFF
--- a/src/Facebook/HttpClients/FacebookCurl.php
+++ b/src/Facebook/HttpClients/FacebookCurl.php
@@ -126,6 +126,9 @@ class FacebookCurl
   public function close()
   {
     curl_close($this->curl);
+
+    // closed handle has to be initialized again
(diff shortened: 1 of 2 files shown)


SammyK:
:+1: Nice catch!

mlebkowski:
what’s the release cycle for this kind of changes? will this land in the next stable or should I PR to `master`?

SammyK:
Yeah, master doesn't suffer from this issue since it doesn't use any statics on the CURL HTTP client.

But patches like this would go in a minor release so it would normally get tagged right away. :)

PR #358, opened 2015-02-15
added support for graph-video url

zoharbabin:
for some reason, the new v4 SDK dropped support for the different API sub domains... which meant that video publishing would use the main api rather than the graph-video subdomain.
This minor fix introduces a change to support the video url.

### DIFF
--- a/src/Facebook/FacebookRequest.php
+++ b/src/Facebook/FacebookRequest.php
@@ -52,6 +52,11 @@ class FacebookRequest
   const BASE_GRAPH_URL = 'https://graph.facebook.com';
 
   /**
+   * @const string Graph API URL
+   */
+  const BASE_VIDEO_GRAPH_URL = 'https://graph-video.facebook.com';
+
+  /**
    * @var FacebookSession The session used for this request
    */
   private $session;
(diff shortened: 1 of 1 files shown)


SammyK:
This is a good first step, but depending on the HTTP client you're using you'll get different results.

- Files need to be posted as `multipart/form-data`
    - For curl implementations, you'll need to make use of [CURLFile](http://php.net/manual/en/class.curlfile.php) class to get the proper encoding.
    - Stream wrappers will break because they [always assume `application/x-www-form-urlencoded` encoding](https://github.com/facebook/facebook-php-sdk-v4/blob/83e96965e4f4591d021b7152b59660bc4e6224bd/src/Facebook/HttpClients/FacebookStreamHttpClient.php#L119).
    - Guzzle should be able to detect the encoding properly if I remember correctly. You might have to explicitly define `multipart/form-data` though.
- The [timeouts of 60 seconds](https://github.com/facebook/facebook-php-sdk-v4/blob/83e96965e4f4591d021b7152b59660bc4e6224bd/src/Facebook/HttpClients/FacebookStreamHttpClient.php#L105) are far too low for video uploads. This would need to be updated on all the HTTP client implementations.

Feel free to port over any of the code from #279 which added file upload support to 4.1.

gfosco:
I don't mind adding this to 4.0, especially since 4.1 isn't released yet..  If the naive approach here doesn't work, for the reasons Sammy mentions, then 4.1 is the answer.

Issue #327, opened 2015-01-23
FacebookCurlHttpClient's call of init() steps on curl opts set manually


jkrasnoo:
Hey all,
In FacebookCurlHttpClient the function openConnection has a call to FacebookCurl's init() before setting options. If a programmer sets a custom FacebookCurl instance within his instantiated client that call to init() will step on that custom FacebookCurl by creating a new handle. The init function doesn't check to see if a handle has already been created.

I needed a custom handle since there really is no other way to implement the proxy settings required to make a connection. Going through the code there is no way to set a proxy for a request. The fix for this is simple and I'll be submitting a pull request.

gfosco:
Assuming this was covered by your PR #328 

Thanks!

PR #359, opened 2015-02-15
Static Code Analysis warnings fixed (partially)

kalessil:
Code cleanup (actually code base is in very good state): code style, if conditionals, strict types, performance. Tests are green.

Following tools are used for code analysis: PhpStorm + <a href="https://plugins.jetbrains.com/plugin/7622?pr=phpStorm">Php Inspections (EA Extended)</a>.

### DIFF
--- a/src/Facebook/Entities/AccessToken.php
+++ b/src/Facebook/Entities/AccessToken.php
@@ -154,11 +154,10 @@ class AccessToken
     $machineIdIsValid = $tokenInfo->getProperty('machine_id') == $machineId;
     $accessTokenIsValid = $tokenInfo->isValid();
 
+    $accessTokenIsStillAlive = true;
     // Not all access tokens return an expiration. E.g. an app access token.
     if ($tokenInfo->getExpiresAt() instanceof \DateTime) {
       $accessTokenIsStillAlive = $tokenInfo->getExpiresAt()->getTimestamp() >= time();
-    } else {
-      $accessTokenIsStillAlive = true;
     }
 
     return $appIdIsValid && $machineIdIsValid && $accessTokenIsValid && $accessTokenIsStillAlive;
@@ -178,7 +177,7 @@ class AccessToken
   {
     $params = array(
       'code' => $code,
-      'redirect_uri' => '',
+      'redirect_uri' => ''
     );
 
     if ($machineId) {
@@ -203,7 +202,7 @@ class AccessToken
 
     $params = array(
       'access_token' => $accessToken,
-      'redirect_uri' => '',
+      'redirect_uri' => ''
     );
 
     return static::requestCode($params, $appId, $appSecret);
@@ -221,7 +220,7 @@ class AccessToken
   {
     $params = array(
       'grant_type' => 'fb_exchange_token',
-      'fb_exchange_token' => $this->accessToken,
+      'fb_exchange_token' => $this->accessToken
     );
 
     return static::requestAccessToken($params, $appId, $appSecret);
@@ -374,7 +373,7 @@ class AccessToken
    */
   public function isAppSession()
   {
-    return strpos($this->accessToken, "|") !== false;
+    return strpos($this->accessToken, '|') !== false;
   }
 
 }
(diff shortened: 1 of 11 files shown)


src/Facebook/Entities/AccessToken.php
```diff
@@ -178,7 +177,7 @@ public static function getAccessTokenFromCode($code, $appId = null, $appSecret =
   {
     $params = array(
       'code' => $code,
-      'redirect_uri' => '',
```
SammyK: I'm :-1: for removing trailing commas from arrays. It doesn't cause any errors and is super convenient when adding new items. One of the nice features of PHP that JavaScript got wrong. :)

src/Facebook/FacebookRequest.php
```diff
@@ -194,13 +194,13 @@ public function __construct(
...
-      && !isset($params["access_token"])) {
-      $params["access_token"] = $session->getToken();
+      && !isset($params['access_token'])) {
+      $params['access_token'] = $session->getToken();
     }
-    if (FacebookSession::useAppSecretProof()
-      && !isset($params["appsecret_proof"])) {
-      $params["appsecret_proof"] = $this->getAppSecretProof(
-        $params["access_token"]
+    if (!isset($params['appsecret_proof'])
```
SammyK: Space between `!` and `isset` so it's more obvious it's negated. And it's still PSR-2 compliant.

src/Facebook/FacebookRequest.php
```diff
@@ -249,7 +249,7 @@ public function execute()
 
     static::$requestCount++;
 
-    $etagHit = 304 == $connection->getResponseHttpStatusCode();
+    $etagHit = (304 === $connection->getResponseHttpStatusCode());
```
SammyK: The parentheses around this statement is redundant and harder to read so it should be like it was before. The strict `===` checking is OK though.

src/Facebook/FacebookRequestException.php
```diff
@@ -80,7 +80,7 @@ public static function create($raw, $data, $statusCode)
     if (!isset($data['error']['code']) && isset($data['code'])) {
       $data = array('error' => $data);
     }
-    $code = (isset($data['error']['code']) ? $data['error']['code'] : null);
+    $code = (isset($data['error']['code']) ? (int)$data['error']['code'] : null);
```
SammyK: Space between `(int)` and `$data` to be consistent with the rest of the code in the SDK.

src/Facebook/HttpClients/FacebookCurl.php
```diff
@@ -33,9 +33,9 @@ class FacebookCurl
 {
 
   /**
-   * @var resource Curl resource instance
+   * @var resource|null Curl resource instance
    */
-  protected $curl = null;
```
SammyK: There was a recent PR that added this. You'll need to duke it out with the original PR. :) See: #328 & #327.

SammyK:
:+1: except for the things I commented on. :)

src/Facebook/Entities/AccessToken.php
kalessil: And it's also a open door for copy paste =) I will revert this changes nevertheless.

src/Facebook/FacebookRequest.php
kalessil: Right, has sense.

src/Facebook/FacebookRequest.php
kalessil: Will remove, for me it's easier to read with parentheses.

src/Facebook/FacebookRequestException.php
kalessil: Right

src/Facebook/HttpClients/FacebookCurl.php
kalessil: That something interesting, thanks for links.

kalessil:
@SammyK : reverted commas, null assignments, added spaces for casting and !isset.

SammyK:
:+1:

PR #361, opened 2015-02-18
[4.1] "Final" doc updates

SammyK:
This PR covers enough documentation to launch 4.1 as stable!

W00t!!!

I hope you have a good copy editor because I'm sure there's a type-o or seven. Also, I'm not sure I got all the FBMD formatting correct.

There's more stuff we can add to this later but this is a really solid start. :) I will eventually update the `CHANGELOG` as that will be really important as we get on a regular `MAJOR` release cycle. But we don't need that to upgrade to 4.1. :)

Once we merge this in & update the docs on Facebook, we can tag 4.1!

@gfosco Balls in your court! :)


yguedidi:
:clap: :)

PR #363, opened 2015-02-19
Fix for PHP Strict Error

haash:
Tested on both PHP 5.4.29 and PHP 5.5.13, the previous code would always return an error of: "Strict Standards: Only variables should be passed by reference in <phpsdk>/src/Facebook/FacebookRequest.php on line 221"

This version fixes the issue and has been tested on PHP 5.4.29.

### DIFF
--- a/src/Facebook/FacebookRequest.php
+++ b/src/Facebook/FacebookRequest.php
@@ -218,7 +218,8 @@ class FacebookRequest
    */
   protected function getRequestURL()
   {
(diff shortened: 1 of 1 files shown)


Loxo:
See https://github.com/facebook/facebook-php-sdk-v4/pull/360
And you got a syntax error.

haash:
@Loxo I guess that's the problem with not copying an pasting. Apologies for duplicating the pull, must have missed it in my glances over the requests. I'll lock this.

PR #364, opened 2015-02-20
.editorconfig added

kalessil:
Simplifies contribution (tabs vs spaces, 4 vs 2). File recognized by some IDEs automatically.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,10 @@
+; top-most EditorConfig file
+root = true
+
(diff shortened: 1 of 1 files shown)


gfosco:
I appreciate the thought, but as 4.0-dev is pretty much done, I'd rather not add this right now.  Future development beyond bug fixes should be targeted at 4.1+.  The master branch is also switched to PSR-2 (4 spaces indent).  Thanks.

PR #365, opened 2015-02-20
Fix for auto-load prefix

niraj-shah:
The extra ```\``` is causing the ```strncmp``` check to fail, resulting in Fatal Errors.


src/Facebook/autoload.php
```diff
@@ -42,9 +42,10 @@
  *
  * @return void
  */
-spl_autoload_register(function ($class) {
+spl_autoload_register(function ($class)
+{
```
yguedidi: This is wrong

yguedidi:
Can you squash your branch in one commit please?

PR #366, opened 2015-02-20
Fixed auto-load prefix

niraj-shah:
The extra ```\``` is causing the ```strncmp``` check to fail, resulting in Fatal Errors.


yguedidi:
:+1:

PR #367, opened 2015-02-24
revert interface change

mlebkowski:
You made a breaking change to an interface in a minor patch library fix (`4.0.16` → `4.0.17`). This fucks up any library implementing this interface. According to [semantic versioning](http://semver.org) this should be a major version bump, so I suggest reverting this change.

### DIFF
--- a/src/Facebook/HttpClients/FacebookCurlHttpClient.php
+++ b/src/Facebook/HttpClients/FacebookCurlHttpClient.php
@@ -141,7 +141,7 @@ class FacebookCurlHttpClient implements FacebookHttpable
    *
    * @throws \Facebook\FacebookSDKException
    */
-  public function send($url, $method = 'GET', array $parameters = array())
+  public function send($url, $method = 'GET', $parameters = array())
   {
     $this->openConnection($url, $method, $parameters);
     $this->tryToSendRequest();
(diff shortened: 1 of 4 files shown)


gfosco:
That's really a breaking change?..  I don't mind removing this type hinting, but what did it break?

SammyK:
Ah yes, the `send()` method signature the `FacebookHttpable` interface changed so any class coded to that interface will now have a different method signature and will cause errors so it's def a BC. Nice catch @mlebkowski!

yguedidi:
@gfosco yes it's a BC break

yguedidi:
@SammyK I'm late lol!

SammyK:
Beat you to it by seconds! :D

gfosco:
Haven't run into this before with a tagged release... Should I tag this 4.0.18 or delete and re-tag 4.0.17?

SammyK:
I'd just tag 4.0.18 and count 4.0.17 as a bug release.

gfosco:
Moved too quickly, it's tagged 4.0.18 but the readme badge and the version string are still .17... (tableflip)

SammyK:
Lol

gfosco:
Fixed.  Thanks all.

yguedidi:
:+1:

mlebkowski:
Thanks for a swift reaction! :D

PR #368, opened 2015-02-24
Update UrandomPseudoRandomStringGenerator.php

sarciszewski:
Prevent `fread()` from wastefully buffering 8192 bytes from `/dev/urandom`.


SammyK:
:+1:

PR #372, opened 2015-03-02
Updated FacebookRedirectLoginHelper and SignedRequest ...

gfosco:
to constant time CSRF check.

### DIFF
--- a/src/Facebook/Entities/SignedRequest.php
+++ b/src/Facebook/Entities/SignedRequest.php
@@ -324,8 +324,18 @@ class SignedRequest
    */
   public static function validateCsrf(array $data, $state)
   {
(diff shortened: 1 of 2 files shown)


src/Facebook/Entities/SignedRequest.php
```diff
@@ -324,8 +324,18 @@ public static function validateSignature($hashedSig, $sig)
    */
   public static function validateCsrf(array $data, $state)
   {
-    if (isset($data['state']) && $data['state'] === $state) {
-      return;
+    if (isset($data['state'])) {
+      $savedLen = mb_strlen($state);
+      $givenLen = mb_strlen($data['state']);
```
yguedidi: Just use `strlen`

gfosco:
Thanks @yguedidi

PR #373, opened 2015-03-02
fix: validate state

ilanco:
After updating to 4.0.19, the isValidRedirect function didn't work right. The wrong variables were checked against each other.

### DIFF
--- a/src/Facebook/FacebookRedirectLoginHelper.php
+++ b/src/Facebook/FacebookRedirectLoginHelper.php
@@ -174,7 +174,6 @@ class FacebookRedirectLoginHelper
    */
   public function getSessionFromRedirect()
   {
-    $this->loadState();
     if ($this->isValidRedirect()) {
       $params = array(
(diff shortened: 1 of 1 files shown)


SammyK:
Oh snap! Nice catch! :)

src/Facebook/FacebookRedirectLoginHelper.php
```diff
@@ -203,7 +203,7 @@ public function getSessionFromRedirect()
    */
   protected function isValidRedirect()
   {
-    $savedState = $this->getCode();
+    $savedState = $this->state;
```
SammyK: This should be `$savedState = $this->loadState();`

ilanco:
Removed the `loadState()` function call from `getSessionFromRedirect()` since it's being called in 
`isValidRedirect()` now.

gfosco:
Merged and tagged 4.0.20.  Thank you

yguedidi:
:+1:

PR #374, opened 2015-03-05
[4.1] Fix HHVM bug

SammyK:
The `stream_set_read_buffer()` function is [not supported on HHVM](http://docs.hhvm.com/manual/en/function.stream-set-read-buffer.php) and is causing the [tests to fail on Travis-CI](https://travis-ci.org/facebook/facebook-php-sdk-v4/jobs/52794560#L180). This is the fix. :)


yguedidi:
"[4.1" here too :)

SammyK:
Oh my goodness. What's wrong with me? Lol.

PR #375, opened 2015-03-05
[4.1] Add Travis-CI check for PHP 7.0 & other changes

SammyK:
- Added Travis-CI check for PHP 7.0 since that's [gonna be thing soon](https://wiki.php.net/rfc/php7timeline#proposal). I've got `allow_failures` on so it won't cause the whole build to fail.
- Removed `allow_failures` for HHVM (because hey - you're Facebook and your PHP SDK should run on your flavor of PHP!) :P
- Added `travis_retry` wrapper to `composer` commands so that it'll [retry on timeouts](http://docs.travis-ci.com/user/build-timeouts/).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,20 +4,21 @@ php:
   - 5.4
   - 5.5
   - 5.6
+  - 7.0
   - hhvm
 
 sudo: false
 
 before_install:
-  - composer self-update
+  - travis_retry composer self-update
 
 install:
-  - composer install --prefer-source --no-interaction
+  - travis_retry composer install --prefer-source --no-interaction
 
 script:
   - vendor/bin/phpunit --coverage-text --exclude-group integration
 
 matrix:
   allow_failures:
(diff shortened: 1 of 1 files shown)


SammyK:
PS: This build will fail since #374 is a fix for the HHVM bug. Once that gets merged in this will pass again. :)

yguedidi:
:+1:, just label the title with "[4.1]" ;)

gfosco:
And the current PHP7 failure is in Guzzle. :)

Cool... now if only I had time to load all the docs.

SammyK:
> And the current PHP7 failure is in Guzzle. :)

Yep! And I let [Michael know](https://github.com/guzzle/guzzle/pull/994). :)

> now if only I had time to load all the docs.

DOOOOOOOOOOEEEEEEEEEETTTT! Hehehe. If you don't get it done before F8 maybe we can sit down and work on it together when I get there. :)

PR #377, opened 2015-03-09
Use bin-dir configuration of composer

matthieuauger:
Composer allows to set up a bin-dir configuration for [vendor binaries](https://getcomposer.org/doc/articles/vendor-binaries.md#can-vendor-binaries-be-installed-somewhere-other-than-vendor-bin-).

I find it easier to run binaries when they are at the root of the project (instead of being buried in the vendor/ directory).

Thus, the tests can be run with just `bin/phpunit` instead of `vendor/bin/phpunit`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,5 @@ composer.phar
 .DS_Store
 .idea/
 tests/FacebookTestCredentials.php
+bin/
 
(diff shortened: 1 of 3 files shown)


gfosco:
Sorry... I've previously denied this change, and as 4.0-dev is just about to be a maintenance release, I'm inclined to keep it the way it is.

If you want to make this change to the master branch (4.1) if it is not already, I'd accept that.

matthieuauger:
No problem, I completely understand that.

I'll check if the the change is present on 4.1

Thanks

Issue #379, opened 2015-03-23
Facebook\Facebook -> delete() should have a "params" parameter


txptr:
Hey everybody,

recently I wondered, why the delete method in the Facebook-Class doesn't have a "params" parameter? For some Facebook Delete-Requests you need the params field to send 
additional data (e.g. unblocking a user - https://developers.facebook.com/docs/graph-api/reference/v2.2/page/blocked).

What do you think?

Greetings

SammyK:
You are correct! It's seems to be fairly uncommon for `DELETE` requests to have an entity body, but if the Graph API supports it, so should the SDK.

Do you want to submit a PR for this & update the docs? :)

txptr:
I hope I did everything right so far?!

gfosco:
Merged #380

PR #380, opened 2015-03-24
Updated delete method to support parameters - #379

txptr:
The Graph API supports parameters for DELETE Requests.
For example: Unblocking a user from a page requires a parameter "user"
which the current version of the SDK doesn't implement.

Docs were updated also.

### DIFF
--- a/src/Facebook/Facebook.php
+++ b/src/Facebook/Facebook.php
@@ -405,6 +405,7 @@ class Facebook
      * Sends a DELETE request to Graph and returns the result.
      *
      * @param string                  $endpoint
+	 * @param array                   $params
      * @param AccessToken|string|null $accessToken
      * @param string|null             $eTag
      * @param string|null             $graphVersion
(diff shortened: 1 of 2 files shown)


src/Facebook/Facebook.php
```diff
@@ -405,6 +405,7 @@ public function post($endpoint, array $params = [], $accessToken = null, $eTag =
      * Sends a DELETE request to Graph and returns the result.
      *
      * @param string                  $endpoint
+	 * @param array                   $params
```
yguedidi: spaces indentation, not tabulation

txptr: I will fix it later...

SammyK:
This should be slated for 4.2 since 4.1 had a feature freeze. :)

gfosco:
Well... since I haven't loaded the docs yet.. lets get this in.

SammyK:
It's technically a BC but it's also a bug so that's cool if you want to slate it for 4.1 :)

gfosco:
Didn't realize it was a BC, but, I'm doing it anyway since this was all @dev... Sorry!

yguedidi:
The is not really a BC break, we don't have any 4.1 release yet :)

Issue #392, opened 2015-04-03
4.0.22,  Fatal error: Cannot use object of type stdClass as array


yiliaofan:
Fatal error: Cannot use object of type stdClass as array in /xxxx/third_party/Facebook/FacebookRedirectLoginHelper.php on line 191

solve the problem:
      if ($response instanceof \stdClass && isset($response->access_token)) {
          return new FacebookSession($response->access_token);
      }
      if (is_array($response) && isset($response['access_token'])) {
        return new FacebookSession($response['access_token']);
      }

SammyK:
Just submitted a fix for this: #393

Issue #391, opened 2015-04-03
Please fix getSessionFromRedirect()


clbphanmem:
We're using previous version of facebook api normally. But when i download the newest Version and run it on old source.It has error : 

"Fatal error: Cannot use object of type stdClass as array in C:\xampp\htdocs\fb2\Facebook\FacebookRedirectLoginHelper.php on line 191" 

Because FacebookRedirectLoginHelper on line 191 uses array and response is Object. I must convert Response into array and it run normally. Please fix or announce something about this issue. :(

Thanks.

xmrflipflop:
Using 4.0.22, experiencing same things, what was the last working version? 
Suggest fix for line 191-193:
      if (isset($response->access_token)) {
        return new FacebookSession($response->access_token);
      }

clbphanmem:
Thank you for responding so quickly. :+1: 
I have another way to fix it. I converted $respond to array, but i like your suggestion because it optimizer than my idea.

SammyK:
Just submitted a fix for this: #393

PR #393, opened 2015-04-03
[4.0][Fix] Fix stdClass error on FacebookRedirectLoginHelper

SammyK:
In Graph `v2.2` and below, GET requests to the `/oauth/access_token` endpoint would return a query string in the body of the response:

```
access_token=foo_token&expires=5183759
```

Starting in Graph `v2.3`, the response body is json:

```
{"access_token":"foo_token","token_type":"bearer","expires_in":5183686}
```

Since v4.0 of the PHP SDK translated these two responses differently, it was causing errors for anyone using Graph v2.3 (#391, #392 and even [here](https://github.com/SammyK/LaravelFacebookSdk/issues/56)). This PR is an ugly hack to fix this without BC's. :)

Not sure if this response change will affect any other places in 4.0, but if we get any more issues about `stdClass` errors, you can bet it's due to the Graph response change.

Version 4.1 of the SDK is unaffected.

### DIFF
--- a/src/Facebook/FacebookRedirectLoginHelper.php
+++ b/src/Facebook/FacebookRedirectLoginHelper.php
@@ -188,8 +188,17 @@ class FacebookRedirectLoginHelper
         '/oauth/access_token',
         $params
       ))->execute()->getResponse();
-      if (isset($response['access_token'])) {
-        return new FacebookSession($response['access_token']);
+
+      // Graph v2.3 and greater return objects on the /oauth/access_token endpoint
+      $accessToken = null;
+      if (is_object($response) && isset($response->access_token)) {
+        $accessToken = $response->access_token;
+      } elseif (is_array($response) && isset($response['access_token'])) {
+        $accessToken = $response['access_token'];
+      }
+
+      if (isset($accessToken)) {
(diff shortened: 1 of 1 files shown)


src/Facebook/FacebookRedirectLoginHelper.php
```diff
@@ -188,8 +188,17 @@ public function getSessionFromRedirect()
         '/oauth/access_token',
         $params
       ))->execute()->getResponse();
-      if (isset($response['access_token'])) {
-        return new FacebookSession($response['access_token']);
+
+      // Graph v2.3 and greater return objects on the /oauth/access_token endpoint
+      $accessToken = null;
+      if ($response instanceof \stdClass && isset($response->access_token)) {
```
yguedidi: `is_object($response)` :)

yguedidi: Maybe you can also use `property_exists($response, 'access_token')`, maybe, I'm +0 about that haha

SammyK:
Thanks @yguedidi! Rebased. :)

goldsatish:
god job keep it up! yes i updated code then after working...

PR #396, opened 2015-04-07
Update FacebookRedirectLoginHelper.php

EliuFlorez:
Change IsSet() To is_null()

### DIFF
--- a/src/Facebook/FacebookRedirectLoginHelper.php
+++ b/src/Facebook/FacebookRedirectLoginHelper.php
@@ -197,11 +197,11 @@ class FacebookRedirectLoginHelper
         $accessToken = $response['access_token'];
       }
 
-      if (isset($accessToken)) {
+      if (is_null($accessToken)) {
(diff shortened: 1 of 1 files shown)


EliuFlorez:
@facebook-github-bot thank you! ready.

SammyK:
Hey @EliuFlorez! Why did you make this change? It breaks the logic.

EliuFlorez:
@SammyK $ accessToken is valid for isset but not for is_null optionally return the null value

yguedidi:
@EliuFlorez I think you mean `!is_null()`, not `is_null()`.
But `isset()` also returns `false` when the value is `null`, even if a variable is "set".

> isset — Determine if a variable is set and is not NULL ([from the doc](http://php.net/manual/en/function.isset.php))

EliuFlorez:
@yguedidi true it is! is_null () :(

yguedidi:
@EliuFlorez so this can be closed, as your change already handled by `isset()` :)

PR #417, opened 2015-05-06
Remove branch alias on 4.1-dev branch

SammyK:
This is a tandem PR with the one I'm about to submit on `master`.

@yguedidi Can you review this one and the one I'm about to submit to ensure this is going to not break the existing aliases and also enable `~5.0@dev`?


yguedidi:
:+1:

PR #418, opened 2015-05-06
Add branch alias for PHP SDK v5

SammyK:
This should add an alias for `~5.0@dev` on the master branch.

@yguedidi Can you ensure this + #417 will do what we've been discussing? :)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,7 +35,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "4.1.x-dev"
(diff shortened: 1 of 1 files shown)


SammyK:
@yguedidi Also, should we change the this line in composer.json:

```
"name": "facebook/php-sdk-v4"
```

To: 

```
"name": "facebook/sdk"
```

And then resubmit the package on packagist under the new namespace? Or how's the best way to do that?

yguedidi:
:+1:

PR #434, opened 2015-05-17
Refactor the GraphAlbum subclass

devmsh:
1. reorder the current fields based on metadata=1
2. add the missing fields
3. remove from fields from the graphObjectMap, it could be user or page,
so it must stay as GraphNode.
4. add the event Field to the graphObjectMap
5. update the description from the metadata
6. refactor the GraphAlbumTest
7. update the GraphNode documentation

### DIFF
--- a/docs/GraphNode.fbmd
+++ b/docs/GraphNode.fbmd
@@ -280,105 +280,135 @@ FB(devsite:markdown-wiki:table {
   columns: ['Property','GraphNode subtype',],
   rows: [
     [
-      '`from`',
-      '[`Facebook\GraphNodes\GraphUser`](#user-instance-methods)',
+      '`event`',
+      '[`Facebook\GraphNodes\GraphEvent`](#event-instance-methods)',
     ],
     [
       '`place`',
       '[`Facebook\GraphNodes\GraphPage`](#page-instance-methods)',
     ],
+    [
+      '`picture`',
+      '[`Facebook\GraphNodes\GraphPicture`](#picture-instance-methods)',
+    ],
   ],
 })
 
-All getter methods return `null` if the property does not exist on the node.
+All getter methods return `null` if the field does not exist on the node.
 
 ### getId() {#album-getid}
 ~~~~
 public string|null getId()
 ~~~~
-Returns the `id` property for the album as a string if present.
-
-### getName() {#album-getname}
-~~~~
-public string|null getName()
-~~~~
-Returns the `name` property for the album as a string if present.
+The album ID.
 
 ### getCanUpload() {#album-getcanupload}
 ~~~~
 public boolean|null getCanUpload()
 ~~~~
-Returns the `can_upload` property for the album as a boolean if present.
+Whether the viewer can upload photos to this album.
 
 ### getCount() {#album-getcount}
 ~~~~
 public int|null getCount()
 ~~~~
-Returns the `count` property for the album as an integer if present.
+Number of photos in this album.
 
 ### getCoverPhoto() {#album-getcoverphoto}
 ~~~~
 public string|null getCoverPhoto()
 ~~~~
-Returns the `cover_photo` property for the album as a string if present.
+Album cover photo id.
 
 ### getCreatedTime() {#album-getcreatedtime}
 ~~~~
 public \DateTime|null getCreatedTime()
 ~~~~
-Returns the `created_time` property for the album as a `\DateTime` if present.
-
-### getUpdatedTime() {#album-getupdatedtime}
-~~~~
-public \DateTime|null getUpdatedTime()
-~~~~
-Returns the `updated_time` property for the album as a `\DateTime` if present.
+The time the album was initially created.
 
 ### getDescription() {#album-getdescription}
 ~~~~
 public string|null getDescription()
 ~~~~
-Returns the `description` property for the album as a string if present.
+The description of the album.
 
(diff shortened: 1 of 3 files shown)


devmsh:
@SammyK  waiting your review

docs/GraphNode.fbmd
```diff
@@ -270,111 +270,116 @@ Returns the `perms` property for the page as an `array` if present. (Only availa
 <card>
 ## GraphAlbum Instance Methods {#album-instance-methods}
 
-The `GraphAlbum` collection represents an [Album](https://developers.facebook.com/docs/graph-api/reference/album) Graph node.
```
SammyK: Is there a reason why you removed this line? I think we should keep the reference to the Graph API docs for this node.

src/Facebook/GraphNodes/GraphAlbum.php
```diff
@@ -28,19 +28,19 @@
...
  */
-
 class GraphAlbum extends GraphNode
 {
     /**
-     * @var array Maps object key names to Graph object types.
+     * @var array Maps object key names to GraphNode types.
      */
     protected static $graphObjectMap = [
-        'from' => '\Facebook\GraphNodes\GraphUser',
```
SammyK: This is technically a BC. Probably not likely someone is really relying on getting a `GraphUser` in the `from` field though. And it's certainly being cast as the wrong subtype as a `GraphUser` so it should be changed. @yguedidi Thoughts?

src/Facebook/GraphNodes/GraphAlbum.php
```diff
@@ -90,17 +90,7 @@ public function getCreatedTime()
...
-     * @return \DateTime|null
-     */
-    public function getUpdatedTime()
-    {
-        return $this->getField('updated_time');
-    }
-
-    /**
-     * Returns the description of the album.
+     * Returns the `description` (The description of the album) as string if present.
```
SammyK: I feel like we should normalize the docblocks on all the getter methods across the board.

I think they should just be the description returned from `metadata=1`.

So this:

    Returns the `description` (The description of the album) as string if present.

Would just be

    The description of the album.

A curious dev just has to look down 2 lines to see what datatype it's being returned as and 2 more lines to see the name of the field. So it'd be nice to not have to maintain all that stuff in both the docblocks and actual code to reduce complexity.

SammyK:
Nice work! Left some feedback and then I'm :+1:  :)

src/Facebook/GraphNodes/GraphAlbum.php
yguedidi: @SammyK It's a BC break, and see #436 for my thoughts about BC ;)

src/Facebook/GraphNodes/GraphAlbum.php
yguedidi: :+1:

docs/GraphNode.fbmd
devmsh: I add this line to my generator, and I also update `GraphEvent`, `GraphLocation`, `GraphPicture`, `GraphAchievement` collections.

src/Facebook/GraphNodes/GraphAlbum.php
devmsh: mmmm, so we need to return the `form` auto cast and add todo comment to delete it later in a the next major release?

src/Facebook/GraphNodes/GraphAlbum.php
devmsh: take a look about the class and the docs now, I think it is more clear and readable now

what are you think?

src/Facebook/GraphNodes/GraphAlbum.php
yguedidi: @devmsh it will depends on #436 final thoughts :)

yguedidi:
Much better, :+1:

src/Facebook/GraphNodes/GraphAlbum.php
SammyK: Yeah, that's hella better! We need to do this for all the other subtypes. :)

docs/GraphNode.fbmd
SammyK: You have a generator to generate the subtypes? Is it open sourced? :)

devmsh: I save the JSON returned from `whatever?fields=id&metadata=1` in a whatever.json file, and use it as an input for my simple generator file (not published because it fully of if else bugs @#$% codes :D

Any way, you can take a look about it in this [Gist](https://gist.github.com/devmsh/81363a68fb62ccc28fb5#file-graph-api-subclass-generator)

I hate the repetitive work, and you hate magic, so I do my own magic behind the seen :P

SammyK: Nice! :)

devmsh:
So, I will wait for #436 and the map refracting, then we can continue with this sub class and other sub classes

gfosco:
Did we want to put this in 5.0 or no?

SammyK:
5.0 or 5.1 - up to @devmsh since there are merge conflicts. :)

devmsh:
I prefere to wait for #437 :)

gfosco:
I'm going to close this.. once 437 is resolved you can submit again.  Thanks!

PR #439, opened 2015-05-21
Fixes #438

### LINKED ISSUE #438: [6.0] Batch Request support for omit_response_on_success
Support should be added for omit_response_on_success

I propose the the following syntax:

``` php
  $requests = [
    'get-friends' => $fb->request('GET', 'me/friends?limit=5', ['omit_response_on_success' => false]),
    'friends' => $fb->request('GET', '?ids={result=get-friends:$.data.*.id}')
  ];

$responses = $fb->sendBatchRequest($requests)->getResponses();
```

sudsy:
I have not written a test for this nor have I added it to the docs

If I am not mistaken, this will submit omit_response_on_success as part of both the query string and as a field in the form data. since it is ignored as a query string parameter by the restful api the solution works.

An ideal solution might remove the parameter from the request object prior to the getBody() request

### DIFF
--- a/src/Facebook/FacebookBatchRequest.php
+++ b/src/Facebook/FacebookBatchRequest.php
@@ -252,7 +252,11 @@ class FacebookBatchRequest extends FacebookRequest implements IteratorAggregate,
             $batch['attached_files'] = $attachedFiles;
         }
 
-        // @TODO Add support for "omit_response_on_success"
+        $requestparams = $request->getParams();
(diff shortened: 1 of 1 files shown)


SammyK:
Thanks for the PR @sudsy!

See my comments in #438. Also check out the [contributing guidelines](https://github.com/facebook/facebook-php-sdk-v4/blob/master/CONTRIBUTING.md) to ensure your PR can be pulled in.

PR #443, opened 2015-05-21
[WIP] Remove Collection

### LINKED ISSUE #440: [6.0][Proposal] Remove Collection class
A node is not a collection, an edge is.

Node:
- Counting fields of a node don't makes sense
- Traversing its fields too
- (ArrayAccess or not is not part of this issue)

Edge:
- Counting number of nodes of an edge makes sense
- Tranversing its nodes too
- ArrayAcces too (don't be surprizied here, I'm not against ArrayAccess, here it's a legitimate usage IMO :) )

I propose to "move" methods that have sense from Collection to GraphEdge and add missing ones to GraphNode

yguedidi:
This closes #440

- [ ] fix tests
- [ ] fix doc

Its just a preview :)


SammyK:
I like where this is going. :)

devmsh:
based on my comment in the original issue, I think that `Collection` class must still there, and only remove the dependency between it and the `GraphNode` class.

While we may have another nodes - open graph node for example - it may be better to have the following:

General `Collection` and `Item` classes
Specific `GraphEdge` and `GraphNode` classes

It will remove the wrong dependencies and still allow for re-usability and will make `GraphNode` and `GraphEdge` code more clear?

gfosco:
Will close this for now.. Our internal tools are not liking my 'leave-issues-open-indefinitely' style.  Re-open as soon as you want.

yguedidi:
:+1: will work on this again once 5.0 is tagged ;)

PR #448, opened 2015-05-30
Missing Forwardslash

gargoyle:
Fixed missing forward slash between version number and the rest of the URL.

### DIFF
--- a/src/Facebook/FacebookRequest.php
+++ b/src/Facebook/FacebookRequest.php
@@ -225,7 +225,8 @@ class FacebookRequest
     } else {
       $baseUrl = static::BASE_GRAPH_URL;
     }
-    return $baseUrl . '/' . $this->version . '/' . $this->path;
+    $normalizedPath = ($this->path[0] == '/') ? $this->path : '/' . $this->path;
+    return $baseUrl . '/' . $this->version . $normalizedPath;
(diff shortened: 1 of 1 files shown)


gfosco:
Paths are provided with the root `/`.  You make a request to `/photos` not just `photos`.  This is at least why it was not in there.  I'm open to reviewing this if you add a function which handles this and allows leaving off the root `/` in request paths.

gargoyle:
Ahh. My bad!

Added a check for a leading `/` and normalize the path accordingly.

src/Facebook/FacebookRequest.php
```diff
@@ -225,7 +225,8 @@ protected function getRequestURL()
     } else {
       $baseUrl = static::BASE_GRAPH_URL;
     }
-    return $baseUrl . '/' . $this->version . $this->path;
+    $normalizedPath = ($this->path[0] == '/') ? $this->path : '/' . $this->path;
```
yguedidi: `return $baseUrl . '/' . $this->version . '/' . ltrim($this->path, '/');`

gargoyle:
Thanks yguedidi - Much neater! :-)

yguedidi:
You're welcome :)

PR #452, opened 2015-06-09
Video uploader - upload by chunks

YuanhuiChen:
Add a video uploader - upload by chunks. Please refer to https://developers.facebook.com/docs/graph-api/video-uploads for the graph api.


gfosco:
Was this based on 4.0-dev or the master branch?  The Travis error makes me think it was based on 4.0-dev..

src/Facebook/FacebookClient.php
```diff
@@ -162,7 +162,11 @@ public function getBaseGraphUrl($postToVideoUrl = false)
      */
     public function prepareRequestMessage(FacebookRequest $request)
     {
-        $postToVideoUrl = $request->containsVideoUploads();
+        $pathElements = explode('/', $request->getUrl());
+        $lastInPath = end($pathElements);
+        $postToVideoUrl = $request->containsVideoUploads() ||
+            ($lastInPath == 'videos' && $request->getMethod() === 'POST');
```
SammyK: Not sure why you added these lines. The `containsVideoUploads()` detects file uploads. We don't want to do any magic guessing what the user is doing based on the endpoint they are poking. :)

src/Facebook/FileUpload/FacebookVideoChunk.php
```diff
@@ -0,0 +1,87 @@
...
+    {
+        parent::__construct($filePath);
+        $this->maxlen = isset($maxlen) ? $maxlen : -1;
+        $this->startOffset = isset($startOffset) ? $startOffset : -1;
+    }
+
+    /**
+     * @param int $startOffset
+     */
+    public function setStartOffset($startOffset)
```
SammyK: This entity should be immutable so I'd kill the setters.

src/Facebook/FileUpload/FacebookVideoUploader.php
```diff
@@ -0,0 +1,255 @@
...
+
+use Facebook\Exceptions\FacebookResponseException;
+use Facebook\Exceptions\FacebookSDKException;
+use Facebook\Facebook;
+
+/**
+ * Class VideoUploadContext
+ * @package Facebook
+ */
+class VideoUploadContext
```
SammyK: The SDK puts all classes in their own file.

src/Facebook/FileUpload/FacebookVideoUploader.php
```diff
@@ -0,0 +1,255 @@
...
+
+    /**
+     * @var int Max retry times
+     */
+    protected $maxTransferTries;
+
+    /**
+     * @var Facebook Reference to Facebook
+     */
+    protected $facebook;
```
SammyK: The Facebook\Facebook super service is a public-facing API and shouldn't be used internally.

src/Facebook/FileUpload/FacebookVideoUploader.php
```diff
@@ -0,0 +1,255 @@
...
+    /**
+     * Upload video file by chunks
+     *
+     * @param string       $filePath
+     * @param int          $targetId
+     * @param string       $accessToken
+     * @param int          $maxTransferTries
+     * @return array
+     */
+    public function upload($filePath, $targetId, $accessToken, $maxTransferTries = 3)
```
SammyK: Since you didn't add any docs, I'm uncertain what the API looks like but I'm curious if there is a slightly better approach.

src/Facebook/FileUpload/FacebookVideoUploader.php
```diff
@@ -0,0 +1,255 @@
...
+
+        return $res;
+    }
+
+    /**
+     * Upload - phase start
+     */
+    public function uploadStart()
+    {
+        $startReqParams = array(
```
SammyK: We use short array syntax throughout the SDK.

src/Facebook/FileUpload/FacebookVideoUploader.php
```diff
@@ -0,0 +1,255 @@
...
+        $this->context->uploadSessionId = $res['upload_session_id'];
+        $this->context->startOffset = $res['start_offset'];
+        $this->context->endOffset = $res['end_offset'];
+        $this->context->videoId = $res['video_id'];
+    }
+
+    /**
+     * Upload - phase transfer
+     */
+    public function uploadTransfer()
```
SammyK: This method is doing too much. It should be broken up into smaller chunks. :)

tests/FacebookVideoUploaderTest.php
```diff
@@ -0,0 +1,108 @@
...
+        $this->testUserId = $res['id'];
+        $this->testUserToken = $res['access_token'];
+    }
+
+    public function testSuccessUploadVideoByChunk()
+    {
+        $uploader = new FacebookVideoUploader($this->testFacebook);
+
+        $res = $uploader->upload(
+            $filePath = __DIR__ . '/foo.mp4',
```
SammyK: Uploading an actual video to Facebook every time we run unit tests is not ideal. :) These tests should mock the responses.

tests/FacebookVideoUploaderTest.php
```diff
@@ -0,0 +1,108 @@
...
+class FacebookVideoUploaderTest extends \PHPUnit_Framework_TestCase
+{
+    protected $testFacebook;
+    protected $testUserId;
+    protected $testUserToken;
+
+    public function setUp()
+    {
+        if (!file_exists(__DIR__ . '/FacebookTestCredentials.php')) {
+            throw new FacebookSDKException(
```
SammyK: These tests fail the Travis-CI checks because they are integration tests. See my comment about mocking the responses. :)

SammyK:
Hey @YuanhuiChen! Thanks for your eagerness to contribute to the SDK!

I left some general feedback throughout your PR. But outside of that, we really need to discuss what the API will look like to upload a video by chucks. Once we all agree on an API, then we can code the implementation. :)

src/Facebook/FacebookClient.php
codingtmd: +1

src/Facebook/FileUpload/FacebookVideoUploader.php
codingtmd: OK. @YuanhuiChen , could we split this videouploadcontext into a separated file under Facebook\FileUpload.

src/Facebook/FileUpload/FacebookVideoUploader.php
codingtmd: Instead of referring Facebook instance directly, could we refer a http client do send the request as 
 https://github.com/facebook/facebook-php-sdk-v4/blob/4.0-dev/src/Facebook/FacebookRequest.php#L92
@SammyK , will that be ok for you?

SammyK: Yes, that's the idea. :)

src/Facebook/FileUpload/FacebookVideoUploader.php
codingtmd: This class is an adapter which receive single file from use, cut it into chunks and feed facebook video upload api. We want to use this class simplify the developer from our chunked upload api(https://developers.facebook.com/docs/graph-api/video-uploads).

I'm not good at doc too. @SammyK , could you give some suggestion about how to improve the doc?


I would suggest to write the doc like this: 
+    /**
+     * Upload video file to Facebook, 
+     *
+     * @param string       $filePath
+     * @param int          $targetId
+     * @param string       $accessToken
+     * @param int          $maxTransferTries
+     * @return array
+     */

src/Facebook/FileUpload/FacebookVideoUploader.php
```diff
@@ -0,0 +1,255 @@
...
+            $res['success'] = false;
+        }
+
+        return $res;
+    }
+
+    /**
+     * Upload - phase start
+     */
+    public function uploadStart()
```
codingtmd: What's the reason to mark this function as public? Maybe it is better to set it as private.

src/Facebook/FileUpload/FacebookVideoUploader.php
SammyK: When adding new features like this, we usually create an issue with the proposed API. Then we all discuss it and decide on the API together. See #438 for a good example of this.

Once the API is decided, then someone codes it and sends a PR. And for new features like this, the PR should contain updates to the documentation explaining how to use it. It's better to keep everyone on the same page. :)

src/Facebook/FileUpload/FacebookVideoUploader.php
codingtmd: @YuanhuiChen ,please set this function as private. Since this function is working on one operation, it might be better to keep things together for readability.

tests/FacebookVideoUploaderTest.php
codingtmd: +1

src/Facebook/FileUpload/FacebookVideoUploader.php
codingtmd: @SammyK , got it. We thought that process is only for external developers. Sorry. Just open an issue for this https://github.com/facebook/facebook-php-sdk-v4/issues/453

Our timeline is to ship this with our July 8 release. Hope we can arrive into an agreed interface soon.
cc @YuanhuiChen

codingtmd:
@gfosco , I remember @YuanhuiChen  told me this diff is based on master, which should be sdk 5.0 if I understand correctly.

gfosco:
Cool, glad it's for 5.0.  When I saw the travis thing about test credentials, I mistakenly assumed it was a 4.0 thing.

src/Facebook/FacebookClient.php
YuanhuiChen: The reason is in the start phase, there is no video files attached, but we need to post to graph-video. 
curl \
  -X POST \
  "https://graph-video.facebook.com/v2.3/1533641336884006/videos"  \
  -F "access_token=XXXXXXXXX" \
  -F "upload_phase=start" \
  -F "file_size=152043520"

src/Facebook/FileUpload/FacebookVideoChunk.php
YuanhuiChen: Sure.

YuanhuiChen:
@gfosco @codingtmd @SammyK 
Thanks for your feedbacks! 
@gfosco This is based on the master branch.

src/Facebook/FacebookClient.php
SammyK: Gotcha. Would love to hear what you think of [my feedback on the API](https://github.com/facebook/facebook-php-sdk-v4/issues/453#issuecomment-110890792). With the API I proposed we wouldn't need to do any parsing-to-guess logic. :)

src/Facebook/FileUpload/FacebookVideoChunk.php
SammyK: That's cool - I'm just saying that every time you update a value you return a new instance of it so all instances are immutable. Just keeps us from depending on a global/shared state. :)

codingtmd:
@SammyK @YuanhuiChen @gfosco , I just add my comments on the API design https://github.com/facebook/facebook-php-sdk-v4/issues/453. Feel free to leave comment there.

And thanks, @SammyK!

YuanhuiChen:
Update the uploader due to the feedbacks and new interface design. Please check! Thanks!

codingtmd:
@SammyK  and @yguedidi , could you guys take a look at the new version? Thanks very much!

src/Facebook/Facebook.php
```diff
@@ -586,4 +589,70 @@ public function videoToUpload($pathToFile)
...
+     * @param int $maxTransferTries
+     *
+     * @return array
+     *
+     * @throws FacebookUploadException
+     */
+    public function uploadVideo($target, $pathToFile, $accessToken, $maxTransferTries = 5)
+    {
+        $uploader = $this->getResumableUploader($this->app, $accessToken, $maxTransferTries);
+        $endpoint = '/'.$target.'/videos';
```
SammyK: We don't want to guess the endpoint for them. They should be able to post to any endpoint.

src/Facebook/Facebook.php
```diff
@@ -586,4 +589,70 @@ public function videoToUpload($pathToFile)
...
+     *
+     * @return array
+     *
+     * @throws FacebookUploadException
+     */
+    public function resumeUpload(FacebookUploadTransferException $exception)
+    {
+        $resumeContext = $exception->getResumeContext();
+
+        $uploader = $this->getResumableUploader($this->app, $resumeContext->accessToken, $resumeContext->maxTransferTries);
```
SammyK: The `getResumableUploader()` method signature does not contain a `FacebookApp` param.

src/Facebook/Facebook.php
```diff
@@ -586,4 +589,70 @@ public function videoToUpload($pathToFile)
...
+        $chunk = $resumeContext->chunk;
```
SammyK: All class properties should be `private` and accessed from getter methods to ensure immutability. This applies to anywhere this is being done. :)

src/Facebook/FacebookClient.php
```diff
@@ -162,7 +162,11 @@ public function getBaseGraphUrl($postToVideoUrl = false)
      */
     public function prepareRequestMessage(FacebookRequest $request)
     {
-        $postToVideoUrl = $request->containsVideoUploads();
+        $pathElements = explode('/', $request->getUrl());
```
SammyK: We still shouldn't try to auto-detect based on the URL here.

src/Facebook/FileUpload/FacebookResumableUploader.php
```diff
@@ -0,0 +1,208 @@
...
+    /**
+     * @param FacebookApp $app
+     * @param string $accessToken
+     * @param int $maxTransferTries
+     */
+    public function __construct(FacebookApp $app, $accessToken, $maxTransferTries = 5)
+    {
+        $this->accessToken = $accessToken;
+        $this->app = $app;
+        $this->client = new FacebookClient();
```
SammyK: This should be injected via the constructor.

src/Facebook/FileUpload/FacebookResumableUploader.php
```diff
@@ -0,0 +1,208 @@
...
+                $filePath,
+                $res['upload_session_id'],
+                $res['video_id'],
+                $res['start_offset'],
+                $res['end_offset'] - $res['start_offset']
+            );
+
+            return $firstTransferChunk;
+        } catch (FacebookResponseException $e) {
+            throw new FacebookUploadException($e->getResponse());
```
SammyK: There should be some logic here to detect the type of error returned. Not all error responses on resumable. Only resumable responses get `FacebookUploadException` (which probably should be named `FacebookResumableUploadException`). For errors that aren't resumable, the `FacebookResponseException` should be re-thrown for the next layer to catch.

src/Facebook/FileUpload/FacebookResumableUploader.php
```diff
@@ -0,0 +1,208 @@
...
+            } catch (FacebookResponseException $e) {
+                if (--$maxTransferTries <= 0) {
+                    $resumeContext = new FacebookResumeContext(
+                        $endpoint,
+                        $this->accessToken,
+                        $chunk
+                    );
+                    throw new FacebookUploadTransferException($e->getResponse(), $resumeContext);
+                }
+                usleep(500000);
```
SammyK: Is this 100 necessary? If so, let's come up with a different method than `usleep()`. This will keep all the resources hogged while it's waiting. Otherwise let's just kill this line.

src/Facebook/FileUpload/FacebookResumableUploader.php
```diff
@@ -0,0 +1,208 @@
...
+     * @throws FacebookUploadTransferException
+     * @throws FacebookSDKException
+     */
+    public function transfer($endpoint, FacebookTransferChunk $chunk)
+    {
+        $transReqParams = [
+            'upload_phase' => 'transfer',
+            'upload_session_id' => $chunk->getUploadSessionId(),
+            'start_offset' => $chunk->getStartOffset(),
+            'video_file_chunk' => $chunk,
```
SammyK: This is sending an object directly into the request params. I think you're wanting to send in the binary chunk of data?

src/Facebook/FileUpload/FacebookResumableUploader.php
```diff
@@ -0,0 +1,208 @@
...
+                $this->app,
+                $this->accessToken,
+                'POST',
+                $endpoint,
+                $finishReqParams
+            );
+
+            $res = $this->client->sendRequest($request)->getDecodedBody();
+            return $res['success'];
+        } catch (FacebookResponseException $e) {
```
SammyK: There should be no catch here since there is not resumable errors on "finish" phase. Let the exception bubble up.

src/Facebook/FileUpload/FacebookResumableUploader.php
SammyK: Sorry - this comment applies to the `transfer` phase. Start phase should let the exception bubble up.

src/Facebook/FileUpload/FacebookTransferChunk.php
```diff
@@ -0,0 +1,116 @@
...
+ *
+ */
+namespace Facebook\FileUpload;
+
+/**
+ * Class FacebookTransferChunk
+ *
+ * @package Facebook
+ */
+class FacebookTransferChunk extends FacebookVideo
```
SammyK: This whole class needs to be refactored a bit and should not extend `FacebookVideo`. @yguedidi Comments on this?

Then `getContents()` on `FacebookVideo` should get two arguments for `$maxLength` & `$offset`

SammyK:
Just did a quick scan and left some feedback. Does the code work yet?

Also - all the new code should have unit tests. And the test video file should be deleted. :)

src/Facebook/Facebook.php
YuanhuiChen: Yes. but in the facebook-php-sdk-v4/src/Facebook/FacebookRequest.php line 197
return $accessTokenEntity->getAppSecretProof($this->app->getSecret());
it requires an app.

src/Facebook/Facebook.php
YuanhuiChen: make sense.

src/Facebook/FacebookClient.php
YuanhuiChen: In the start phase 
  -X POST \
  "https://graph-video.facebook.com/v2.3/1533641336884006/videos"  \
  -F "access_token=XXXXXXXXX" \
  -F "upload_phase=start" \
  -F "file_size=152043520"
there is no video file attached. so it's not safe to determine whether it is video request only by containsVideoUploads.

src/Facebook/FileUpload/FacebookResumableUploader.php
YuanhuiChen: The idea is that we want the user to wait for certain time before next trial. You are right. This maybe not a good way. 
@codingtmd what do you think?

src/Facebook/FileUpload/FacebookResumableUploader.php
YuanhuiChen: It will send binary chunk. The send request client will read the file into binary chunk,  according to the start offset and chunk size.

YuanhuiChen:
Thanks @SammyK  for your detailed review. I still very confused about the unit tests. In the unit test, should I mock request and response or really sending request to facebook? If I mock request, the start, transfer or finish phase will throw out exceptions. There is no place I could insert a mock response. If I post a real request to facebook, then I may need a pho test file. Could you help me with this as well?

src/Facebook/FacebookClient.php
codingtmd: @YuanhuiChen  , only 'transfer' phase has video file attached. 'start' and 'finish' has not video file.

@SammyK , any suggestion for such cases?

src/Facebook/FileUpload/FacebookResumableUploader.php
codingtmd: +1

src/Facebook/FileUpload/FacebookResumableUploader.php
codingtmd: I asked @YuanhuiChen to add this line before.  Let's kill this line for simple.

src/Facebook/FileUpload/FacebookResumableUploader.php
codingtmd: Agree. We only need to catch exception in 'transfer' for resumable purpose.

src/Facebook/FileUpload/FacebookTransferChunk.php
codingtmd: I'm not sure for this part. From naming, it is confusing for a chunk to extend facebookvideo. But these two objects (videofile, chunkfile) has really similar definitions. Creating a separated class for trunk sounds a little heavy. Any other base class we can inherent for chunk purpose?

codingtmd:
@SammyK , could you share a bit about the UT standard in your project? I just go through some tests in the 'test' folder. Looks the test case also post the request to facebook. I don't see the mock for response. Are you looking for integration test or unit test?

It would be great if you can share some sample for @YuanhuiChen to refer.

And thanks a lot for your time to work with us!

SammyK:
> should I mock request and response or really sending request to facebook?

All the responses from the Graph API should be mocked so that we're not making any live requests.

> If I mock request, the start, transfer or finish phase will throw out exceptions.

All of the unit tests should be able to pass independently of each other.

> There is no place I could insert a mock response.

You need to [inject a stub HTTP client](https://github.com/facebook/facebook-php-sdk-v4/blob/master/tests/FacebookClientTest.php#L41-L50) with a fixed response.

> Looks the test case also post the request to facebook.

There are only 3-4 live calls to the Graph API in the existing tests. All the others are mocked. But it'd be nice to eventually get the live calls down to 0. :)

src/Facebook/FileUpload/FacebookResumableUploader.php
codingtmd: @SammyK , we have the error code definition for either resumable or non-resuamable(https://developers.facebook.com/docs/graph-api/video-uploads#errors). We can add the logic to distinguish error type, but need to fine a place to store the error code? Definitely, we can not harcode all resumable error code in the logic. Is there any config file or something else we can take advantage of in the php sdk?  Thanks!

cc @YuanhuiChen

SammyK: I'd just add the resumable error codes to the [error handler](https://github.com/facebook/facebook-php-sdk-v4/blob/5.0-dev/src/Facebook/Exceptions/FacebookResponseException.php#L69-L130) and have it throw a `FacebookResumableUploadException` or something. :)

gfosco:
@codingtmd @YuanhuiChen where are we at with this?

codingtmd:
@YuanhuiChen , could you share an update on this?

codingtmd:
@gfosco  @SammyK , 

@YuanhuiChen just submit another revision for review. Here: https://github.com/facebook/facebook-php-sdk-v4/pull/489

Could you guys take a look and share some feedbacks?

PR #455, opened 2015-06-17
Normalize php documentation in SignedRequest.php

christophlehmann:
- doctrine/annotations is unhappy with @returns


SammyK:
:+1: Thanks! :)

PR #465, opened 2015-07-08
Updates made during load to developers.facebook.com

gfosco:
Formatting updates and changes (and more could be made to make them look better.)


SammyK:
BAM! @gfosco be like

PR #468, opened 2015-07-10
Fix: Method accepts only two arguments, not four

localheinz:
This PR

* [x] removes arguments from a method call as the signature only accepts two anyway


gfosco:
@SammyK Does this mean something else is broken?

SammyK:
Nope - those are dingleberries from when I was pulling out my hair writing that code. I hope someone will eventually rewrite all the graph node auto-casting stuff. It's ugly. :/

gfosco:
Thanks @localheinz :+1:   If you'd like to make the same change against Master, will merge that too.

PR #471, opened 2015-07-19
Fix typo in docs

peterkokot:


gfosco:
Thanks, will also fix in master and on the dev-site!

PR #484, opened 2015-08-04
Make default field value return exact value

### LINKED ISSUE #480: Improve getFields method of Collection.php
The return of the method getFields (Collection.php line 72) should simply be:
```php
return $default
```

Indeed if I call the method (as it is right now) like that:
```php
$tz = $user->getField('timezone', 0);
```
`$tz` will be `null` instead of being equals to `0`

Cheers

SammyK:
Fixes #480. :)


gfosco:
:+1:

PR #489, opened 2015-08-17
Video Resumable uploader

YuanhuiChen:
Finally comes another update.  Sorry I opened a new branch. You cannot see the old comments. I committed to some others' diff by mistake in the old branch. 
The new changes:
1. add integration test
2. distinguish between no resumable exception and resumable exception in the transfer phase.
3. fix some other issues. 

Thanks again for your patience!


gfosco:
This is great, and I'm willing to accept it as is.  I wish it could go against version 5, but I guess that wasn't possible.  Will leave it here for another day to give @SammyK & @yguedidi a chance to comment/nit if they want.

src/Facebook/Facebook.php
```diff
@@ -586,4 +589,72 @@ public function videoToUpload($pathToFile)
...
+     * @param int $maxTransferTries
+     *
+     * @return array
+     *
+     * @throws FacebookSDKException
+     */
+    public function uploadVideo($target, $pathToFile, $accessToken, $maxTransferTries = 5)
+    {
+        $uploader = $this->getResumableUploader($accessToken, $maxTransferTries);
+        $endpoint = '/'.$target.'/videos';
```
SammyK: We shouldn't auto-suffix the endpoint with `videos` since it might not always be that and none of the other parts of the SDK auto-suffix endpoints. :)

src/Facebook/FileUpload/FacebookResumableUploader.php
```diff
@@ -0,0 +1,201 @@
...
+     *
+     * @param $endpoint
+     * @param $filePath
+     * @return FacebookTransferChunk
+     *
+     * @throws FacebookSDKException
+     */
+    public function start($endpoint, $filePath)
+    {
+        $fileSize = filesize($filePath);
```
SammyK: We should add a method to `FacebookFile` called `getSize()` that would do this. And then pass in the instantiation of `FacebookVideo` instead of `$filePath` here. :)

src/Facebook/FileUpload/FacebookResumableUploader.php
```diff
@@ -0,0 +1,201 @@
...
+
+        $request = new FacebookRequest(
+            $this->app,
+            $this->accessToken,
+            'POST',
+            $endpoint,
+            $startReqParams
+        );
+
+        $res= $this->client->sendRequest($request)->getDecodedBody();
```
SammyK: Space after `$res`. Also if `sendRequest()` throws this is going to be problematic as a chained method call.

src/Facebook/FileUpload/FacebookResumableUploader.php
```diff
@@ -0,0 +1,201 @@
...
+     * Upload - phase transfer
+     *
+     * @param $endpoint
+     * @param FacebookTransferChunk $chunk
+     * @return FacebookTransferChunk
+     *
+     * @throws FacebookResumableUploadException
+     * @throws FacebookSDKException
+     */
+    public function transfer($endpoint, FacebookTransferChunk $chunk)
```
SammyK: This method needs to be refactored a bit. Maybe @yguedidi has some thoughts on extracting out some of this logic to other methods. Maybe after it gets merged we can play with refactoring it a bit before tagging it.

tests/FileUpload/FacebookResumableUploaderTest.php
```diff
@@ -0,0 +1,222 @@
...
+        $this->fbApp = new FacebookApp('app_id', 'app_secret');
+        $this->fbSuccessClient = new FacebookClient(new VideoUploadClientHandlerSuccess());
+        $this->fbFailClient = new FacebookClient(new VideoUploadClientHandlerFail());
+    }
+
+    public function testSuccess()
+    {
+        $fileUploader = new FacebookResumableUploader($this->fbApp, $this->fbSuccessClient, 'access_token');
+        $endpoint = '/113582528973300/videos';
+        $filePath = __DIR__ . '/../foo.mp4';
```
SammyK: I still don't think we need to include an actual video file in the SDK. We can mock this file. (Also another reason we need to pass in the `FacebookVideo` instance instead of the path to the file.)

src/Facebook/Facebook.php
```diff
@@ -586,4 +589,72 @@ public function videoToUpload($pathToFile)
...
+
+    /**
+     * Instantiates a new video uploader
+     *
+     * @param string $accessToken
+     * @param int $maxTransferTries
+     *
+     * @return FacebookResumableUploader
+     */
+    protected function getResumableUploader($accessToken, $maxTransferTries = 5)
```
SammyK: Can we rename this method to `getResumableUploadClient`?

src/Facebook/Facebook.php
```diff
@@ -586,4 +589,72 @@ public function videoToUpload($pathToFile)
...
+    {
+        return new FacebookResumableUploader($this->app, $this->client, $accessToken, $maxTransferTries);
```
SammyK: Can we rename this service to `FacebookResumableUploadClient`? :)

SammyK:
I did a quick scan a pointed out a few things. I think after @yguedidi has a look we should do some refactoring after it gets merged before we tag it as 5.1 :)

src/Facebook/Facebook.php
yguedidi: @SammyK uploading a video can only be done with an endpoint formated like `/{node-id}/videos`, so I think it's OK here, end-user will only need to give the node id.

src/Facebook/Facebook.php
```diff
@@ -586,4 +589,72 @@ public function videoToUpload($pathToFile)
...
         return new FacebookVideo($pathToFile);
     }
+
+    /**
+     * Upload video
+     *
+     * @param int $target
+     * @param string $pathToFile
+     * @param string $accessToken
+     * @param int $maxTransferTries
```
yguedidi: Add some parameter descriptions here

src/Facebook/Facebook.php
SammyK: @yguedidi I see what you're saying. I'm also concerned that users will be reading the Graph docs and then try to enter the endpoint directly here. I for one wouldn't assume the SDK would add the `/videos` suffix for me. So I'm still against auto-suffixing. :)

SammyK:
Also - can you add some docs? :)

src/Facebook/Facebook.php
yguedidi: I will just name it `FacebookVideoUploader` :)

SammyK: You would! :) I was just thinking since it's an HTTP client, it'd make more since to make it more descriptive, but I'm cool with what you guys think too :)

src/Facebook/FileUpload/FacebookResumableUploader.php
yguedidi: Why not :)
It's depend how we want this class API be.

src/Facebook/FileUpload/FacebookResumableUploader.php
yguedidi: @SammyK it shouldn't be problematic as a throw stop the execution :)

src/Facebook/FileUpload/FacebookResumableUploader.php
```diff
@@ -0,0 +1,201 @@
...
+            $this->app,
+            $this->accessToken,
+            'POST',
+            $endpoint,
+            $transReqParams
+        );
+
+        while ($maxTransferTries > 0) {
+            try {
+                $res= $this->client->sendRequest($request)->getDecodedBody();
```
yguedidi: space

src/Facebook/FileUpload/FacebookResumableUploader.php
```diff
@@ -0,0 +1,201 @@
...
+
+                $nextTransferChunk = new FacebookTransferChunk(
```
yguedidi: can be returned directly

src/Facebook/FileUpload/FacebookResumableUploader.php
SammyK: @yguedidi Oh true. :)

src/Facebook/FileUpload/FacebookResumableUploader.php
```diff
@@ -0,0 +1,201 @@
...
+                    $chunk->getUploadSessionId(),
+                    $chunk->getVideoId(),
+                    $res['start_offset'],
+                    $res['end_offset'] - $res['start_offset']
+                );
+
+                return $nextTransferChunk;
+            } catch (FacebookSDKException $e) {
+                $preException = $e->getPrevious();
+                if ($preException instanceof FacebookResumableUploadException) {
```
yguedidi: please use an early return here

src/Facebook/FileUpload/FacebookResumableUploader.php
```diff
@@ -0,0 +1,201 @@
...
+                            $this->accessToken,
+                            $chunk
+                        );
+                        $preException->setResumeContext($resumeContext);
+                        throw $e;
+                    }
+                } else {
+                    throw $e;
+                }
+                continue;
```
yguedidi: useless

src/Facebook/Exceptions/FacebookResumableUploadException.php
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    /**
+     * @param FacebookResumeContext $resumeContext
+     */
+    public function setResumeContext(FacebookResumeContext $resumeContext)
+    {
+        $this->resumeContext = $resumeContext;
+    }
+
```
yguedidi: remove this extra empty line

yguedidi: BTW, I don't like the mutability of this class... A first solution that comes to me right now is to merge `FacebookResumeContext` in this exception class. @gfosco @SammyK thoughts?

src/Facebook/Facebook.php
```diff
@@ -45,8 +46,10 @@
 use Facebook\Helpers\FacebookJavaScriptHelper;
 use Facebook\Helpers\FacebookPageTabHelper;
 use Facebook\Helpers\FacebookRedirectLoginHelper;
+use Facebook\Exceptions\FacebookResumableUploadException;
 use Facebook\Exceptions\FacebookSDKException;
 
+
```
yguedidi: should be removed

src/Facebook/Facebook.php
yguedidi: this comment apply to all other public methods :)

src/Facebook/Facebook.php
```diff
@@ -586,4 +589,72 @@ public function videoToUpload($pathToFile)
...
+     * @param int $target
+     * @param string $pathToFile
+     * @param string $accessToken
+     * @param int $maxTransferTries
+     *
+     * @return array
+     *
+     * @throws FacebookSDKException
+     */
+    public function uploadVideo($target, $pathToFile, $accessToken, $maxTransferTries = 5)
```
yguedidi: You should add the Graph version

src/Facebook/Facebook.php
yguedidi: I'd prefer an easy to use signature with a good doc rather an opened signature to any endpoint with a pattern check in the definition :)

src/Facebook/Facebook.php
```diff
@@ -586,4 +589,72 @@ public function videoToUpload($pathToFile)
...
+        $chunk = $uploader->start($endpoint, $pathToFile);
+
+        do {
+            $chunk = $uploader->transfer($endpoint, $chunk);
+        } while(!$chunk->isLastChunk());
+
+        return [
+            'videoId' => $chunk->getVideoId(),
+            'success' => $uploader->finish($endpoint, $chunk->getUploadSessionId()),
+        ];
```
yguedidi: above code starting from the `do {` can be extracted to a private function, as it's used in both `uploadVideo` and `resumeUpload`

src/Facebook/FileUpload/FacebookResumableUploader.php
yguedidi: As a service, this class shouldn't be aware of max transfer tries (a call to `transfer` should just send the request to transfer a chunk).
Then, the max transfer tries part should go in the `Facebook` class, in the `do...while` loop. (and thanks to the code extraction proposed by my above comment, no code duplication)

src/Facebook/FileUpload/FacebookResumableUploader.php
```diff
@@ -0,0 +1,201 @@
...
+        $request = new FacebookRequest(
+            $this->app,
+            $this->accessToken,
+            'POST',
+            $endpoint,
+            $finishReqParams
+        );
+
+        $res = $this->client->sendRequest($request)->getDecodedBody();
+        return $res['success'];
```
yguedidi: add an empty space above the `return`

src/Facebook/FileUpload/FacebookResumableUploader.php
```diff
@@ -0,0 +1,201 @@
...
+
+        $request = new FacebookRequest(
+            $this->app,
+            $this->accessToken,
+            'POST',
+            $endpoint,
+            $finishReqParams
+        );
+
+        $res = $this->client->sendRequest($request)->getDecodedBody();
```
yguedidi: @YuanhuiChen rename all your `$res` to `$response`, `$decodedBody`, `$data`, or any thing else but please, a full word. :)

src/Facebook/Exceptions/FacebookResumableUploadException.php
SammyK: I also don't like mutability. Not really crazy about merging `FacebookResumeContext` either. The `FacebookResumeContext` can be constructed by examining the request & response, right? This might be a bigger architecture issue. Can't really dive too deeply into the PR at the moment but I can make tweaks after a merge and before a tag to review.

src/Facebook/Facebook.php
SammyK: And falling back to the default Graph version:

```php
$graphVersion = $graphVersion ?: $this->defaultGraphVersion;
```

src/Facebook/Exceptions/FacebookResumableUploadException.php
yguedidi: As it's a little bit earlier in the night than yesterday, here a better idea: keep the context class, but slip to two exceptions, a `FacebookFailedUploadException` created in the `FacebookResponseException`, and the `FacebookResumableUploadException` instanciated with the context and throwed when needed, i.e. when we catch the failed exception in the transfer and want to allow an upload retry. :)
BTW, I still think we can have the three little properties of the context (without the max retries) as part of the `FacebookResumableUploadException`: the context and the exception will be instanciated at the same time.

yguedidi:
@gfosco @SammyK I don't like merging "as is" and then refactor and/or fix things. I think when merging a PR, that PR should in a "ready to release" state. Just my 2 cents, will do how @gfosco want as usual :)

src/Facebook/Facebook.php
yguedidi: Yep!

SammyK:
@yguedidi Maybe we should send PR's to @YuanhuiChen's fork?

yguedidi:
@SammyK it's a solution, but are we in a hurry? I think it's better to let @YuanhuiChen finish it, at least to learn more, as code review is an important part of the learning process :)
Of course, if @YuanhuiChen explicitly tell us he/she don't have the time to finish it, we'll help !

SammyK:
@yguedidi No rush. Just thought we could submit some PR's to "talk code" instead of English as we try to figure out the best way to implement this feature since it's such a beast. :) But I'm OK with letting @YuanhuiChen do it without PR's from us as long as he has the time!

yguedidi:
@SammyK sure I talk code better than english haha, but we also can't have some time, you know that too

src/Facebook/Facebook.php
codingtmd: +1. For the chunked upload api, we only allow it with > v2.3. Here, the defaultGraphVersion should be v2.3.

src/Facebook/Facebook.php
codingtmd: Personally, I'm with @yguedidi here. This uploader is designed for uploading videos. If we specify clear that 'target' means user/page id in the doc, user won't pass in a full endpoint path in.

src/Facebook/Facebook.php
codingtmd: +1

src/Facebook/Facebook.php
codingtmd: Either FacebookVideoUploader or  FacebookVideoUploadClient is fine to me.  The reason is, inside the function, we already use the name like 'resumeUpload', which will provide the information to enduser.

@SammyK , your opinion?

src/Facebook/FileUpload/FacebookResumableUploader.php
codingtmd: Not sure whether we are in the same page. I feel that this max tries logic should be moved into the do...while loop of function resumeUpload. @yguedidi, is it what you mean?

src/Facebook/FileUpload/FacebookResumableUploader.php
codingtmd: True. $nextTransferChunk doesn't bring value here. Let's just return it directly in Line 145

src/Facebook/FileUpload/FacebookResumableUploader.php
codingtmd: Not sure I get it. @yguedidi , are you suggesting to refactor the code like this?
```
                  $preException = $e->getPrevious();
                 if (! $preException instanceof FacebookResumableUploadException) {
                     throw $e;
                 }
                 if (--$maxTransferTries <= 0) {
                     $resumeContext = new FacebookResumeContext(
                          $endpoint,
                          $this->accessToken,
                          $chunk
                      );
                      $preException->setResumeContext($resumeContext);
                       throw $e;
                  }
```

tests/FileUpload/FacebookResumableUploaderTest.php
codingtmd: @YuanhuiChen , could we mock this file? from the code, looks we only need to mock one function 'getSize', right?

codingtmd:
Thank you, guys, especially for @YuanhuiChen to work on this task! I just left some comments in diff. We are pretty close, wow!

src/Facebook/FileUpload/FacebookResumableUploader.php
yguedidi: Yes, more precisly in the newly extracted function with the `do...while` loop in it.

src/Facebook/FileUpload/FacebookResumableUploader.php
yguedidi: Exactly! Early returns avoid nesting tabs too much

tests/FileUpload/FacebookResumableUploaderTest.php
SammyK: @codingtmd We just need to mock a `FacebookVideo` and inject it instead of using the path to the file. :)

src/Facebook/Facebook.php
SammyK: I still like naming it a client since it's an HTTP client, but I'm not too fussy about the name. :)

src/Facebook/Facebook.php
SammyK: The chunked video feature was added in 2.3 but I think it's back-versioned to work with all versions of Graph. I can't find the doc reference to that and I haven't tested it myself but I feel like I remember hearing that at F8 this year. :)

On that point though - I think we should allow these types of checks to happen on the Graph API side so we don't have to keep up with all these little "gotchas" on the SDK end as the Graph API changes. Thoughts?

src/Facebook/Facebook.php
yguedidi: @SammyK it is not an HTTP client, it's a video uploader, a helper service that ease resumable video uploading, and it has a dependency on an HTTP client, so I'm still for `FacebookVideoUploader` ;)

src/Facebook/Facebook.php
yguedidi: :+1: to let the Graph API handle it

src/Facebook/Facebook.php
SammyK: @yguedidi I can see your point. There are a number of features of the SDK that make HTTP calls and they are not always so obvious to the end-user that that's what happening in the background (like `getAccessToken()` for example). That's the main reason I suggested the name change, but again, I'm cool with keeping it as-is. :)

src/Facebook/Facebook.php
codingtmd: Graph API has the check for versions. And it will handle the logic differently based on the api version. User can set whatever version they want, but here, for this function, since we use chunked upload, if we want a value for $this->defaultGraphVersion, it should be at least v2.3.

src/Facebook/Facebook.php
codingtmd: Let's just use FacebookVideoUploader.

src/Facebook/Exceptions/FacebookResumableUploadException.php
codingtmd: Sounds a little overkilled. I feel merging FacebookResumeContext into FacebookResumableUploadException makes more sense, since this context is only used with resumableuploadexception. Could we just merge it?

src/Facebook/Facebook.php
SammyK: This won't be an issue since the fallback Graph version in the SDK will always be the latest version. :)

src/Facebook/Exceptions/FacebookResumableUploadException.php
SammyK: Maybe merging is best if `FacebookResumeContext` is only ever used in the case of an exception.

gfosco:
I'm just the shepherd :) ... I am quite happy to see all the discussion and progress, and whenever you're all ready, we'll get 5.1 out. :+1:

gfosco:
@YuanhuiChen will you be able to make the requested changes?   Thanks

codingtmd:
@gfosco, I just heard yesterday that @YuanhuiChen already moved to another high priority task and she will not able to finish this diff. We are short of resource right now. Just wonder, is it possible to find some love from your side on this diff? Otherwise, we need to wait for another two weeks before we can find a new resource.

Cc @SammyK, @yguedidi

SammyK:
I might have some time next week if we can't get anyone else on it. :)

codingtmd:
Great! Thanks, @SammyK.

codingtmd:
Hello @SammyK,  how things going? :)

SammyK:
Hey @codingtmd! Thanks for the ping! I didn't get a chance to look at this last week so I'm going to see what's involved with this tomorrow and let you know. :)

SammyK:
Hey @YuanhuiChen! If I send a PR to [your branch](https://github.com/YuanhuiChen/facebook-php-sdk-v4/tree/video_uploader_chunks) for this feature, will you be able to merge any changes so we can continue the discussion here? :)

codingtmd:
Hello @SammyK, @YuanhuiChen already left this project, so I'm not sure whether she will response or not. Is it possible to send a PR on the master?:)

SammyK:
I think I'll try to create a new branch from @YuanhuiChen's fork so that all the existing commits will stay in place. Things got busy all of a sudden on my end but I'm still trying to work on this when I can. :)

SammyK:
Moved this PR to #498 so we can close this one. :)

PR #491, opened 2015-08-21
added a getEmail() method for GraphUser objects

leonar15:
just adding a helper to fetch the graph user's email address, if any


gfosco:
Ok, will add this.. Thanks :+1:

PR #494, opened 2015-08-27
Add an option to set the app access token as default

pespantelis:
Then you can do this:
```php
$fb = new Facebook\Facebook([
    'default_access_token' => 'app_token'
]);

// or

$fb->setDefaultAccessToken('app_token');
```

### DIFF
--- a/src/Facebook/Facebook.php
+++ b/src/Facebook/Facebook.php
@@ -283,6 +283,12 @@ class Facebook
      */
     public function setDefaultAccessToken($accessToken)
     {
+        if ($accessToken === 'app_token') {
+            $this->defaultAccessToken = $this->app->getAccessToken();
+
+            return;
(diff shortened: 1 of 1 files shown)


pespantelis:
Done.

gfosco:
Interesting.  I'm wondering if this could be a vulnerability.  Could you imagine a case where a developer would call setDefaultAccessToken on something provided by a user?

SammyK:
@gfosco Good call - it's pretty common to log a user in with the JavaScript SDK and then set the default access token in PHP (via URL param or cookie) which could be exploited to contain `app_token`. Then if the app exposes as URL with the access token in somewhere, it will contain the app ID & app secret.

Thanks for the idea @pespantelis! But I'm thinking this could be a possible exploit.

PR #498, opened 2015-09-22
Video upload by chunks

SammyK:
Did most of the changes from the comments in #489. Step two is to refactor all this stuff. :)


codingtmd:
Awesome! @yguedidi , could you also take a look?

By the way, what 'CLA Signed' means?

SammyK:
Thanks @codingtmd. :)

> By the way, what 'CLA Signed' means?

It's just the bot making sure I've signed Facebook's [CLA](https://code.facebook.com/cla). I signed it a while ago. Hehe.

I wanted to make sure we're on the same page with the API before doing any refactoring.

To the end user, they want to upload a video by chucks and we make it as simple as possible:

```php
$fb = new Facebook\Facebook([/* ... */]);

$data = [
  'title' => 'My awesome video',
  'description' => 'More info about my awesome video.',
  'thumb' => $fb->fileToUpload('/path/to/thumbnail.jpg'), // optional
];

$maxRetries = 6;
$graphVersion = 'v2.4'

try {
  $response = $fb->uploadVideo('{foo-id}', '/path/to/video.mp4', $data, '{access-token}', $maxRetries, $graphVersion);
} catch(Facebook\Exceptions\FacebookSDKException $e) {
  echo 'Error: '.$e->getMessage();
  exit;
}

$videoId = $response['video_id'];
```

Does that look about right?

codingtmd:
Got it, @SammyK!

And for your question, yes, your sample is exactly what we expect user to behave.

yguedidi:
:+1: for the API

yguedidi:
@SammyK go for the refactoring, you know where to find my comments :)

SammyK:
*phew* Just refactored. Tested it and it uploads in chunks nicely. It's not my best refactor in the world but it'll do. :) Comments welcome as always.

yguedidi:
I will review it tonight.

SammyK:
The code checks failed - retriggering...

src/Facebook/Facebook.php
```diff
@@ -586,4 +587,38 @@ public function videoToUpload($pathToFile)
...
+     */
+    public function uploadVideo($target, $pathToFile, $metadata = [], $accessToken = null, $maxTransferTries = 5, $graphVersion = null)
+    {
+        $accessToken  = $accessToken ?: $this->defaultAccessToken;
+        $graphVersion = $graphVersion ?: $this->defaultGraphVersion;
+
+        $uploader = new FacebookResumableUploader($this->app, $this->client, $accessToken, $graphVersion);
+        $endpoint = '/'.$target.'/videos';
+        $file     = $this->videoToUpload($pathToFile);
+        $chunk    = $uploader->start($endpoint, $file);
```
yguedidi: I don't like thoses alignment, the reason is that it may make the diff harder to read. Imagine you add a new `$foobarbaz` variable, you will need to change all other line without any added value, and the diff will show multiple lines change but in fact you only add one. :)

yguedidi: This is a personal opinion, I don't really care about the change

src/Facebook/FileUpload/FacebookResumableUploader.php
```diff
@@ -0,0 +1,193 @@
...
+     *
+     * @param string $endpoint
+     * @param FacebookTransferChunk $chunk
+     * @param int $retryCountdown
+     *
+     * @return FacebookTransferChunk
+     *
+     * @throws FacebookResponseException
+     */
+    public function maxTriesTransfer($endpoint, FacebookTransferChunk $chunk, $retryCountdown)
```
yguedidi: I still think that retrying is like a sweet sugar and it shouldn't be part of this service class, it gives it more responsability that just providing function for each phase of a chunk upload.
Having that sweet sugar provided by the `Facebook` master class makes sense, it's an easy way to do for any standard user.
I think it's up to the advanced user that directly use this service to handle retrying an upload.

src/Facebook/Facebook.php
SammyK: Oops! Yeah, my formatter did this when I was doing a quick PSR-2 fix. I don't like it either. :) I'll change it.

src/Facebook/FileUpload/FacebookResumableUploader.php
SammyK: I actually did have this originally in the master `Facebook` class but I was passing around lots of service classes/entities and it seemed like a lot of work of that class. To me it was a smell so I moved it back to the `FacebookResumableUploader`. I don't really see it as syntactic sugar since a resumable uploader seems like a pretty good place to handle retries of a resumable download.

But I'm not 100% married to either way. I'd be happy to move it without too much of a fight. :)

codingtmd: Moving the retry to master does provide the flexibility for every service. 

For chunked upload case, most of the failures happens in transfer. If having this sweet sugar brings too much work, I also feel happy to accept existing implementation of maxTriesTransfer.

SammyK:
Made changes based on feedback. Trying to squash the commits causes issues. But the more I learn about git and see how other people use it, I'm thinking we should stop squashing commits. I've been doing too many force pushes. :)

src/Facebook/Facebook.php
```diff
@@ -586,4 +588,66 @@ public function videoToUpload($pathToFile)
...
+     * @param FacebookResumableUploader $uploader
+     * @param string $endpoint
+     * @param FacebookTransferChunk $chunk
+     * @param int $retryCountdown
+     *
+     * @return FacebookTransferChunk
+     *
+     * @throws FacebookSDKException
+     */
+    public function maxTriesTransfer(FacebookResumableUploader $uploader, $endpoint, FacebookTransferChunk $chunk, $retryCountdown)
```
yguedidi: Should be private

src/Facebook/Facebook.php
```diff
@@ -586,4 +588,66 @@ public function videoToUpload($pathToFile)
...
+    {
+        $allowToThrow = $retryCountdown < 1;
```
yguedidi: Useless variable

src/Facebook/Facebook.php
codingtmd: +1

codingtmd:
Other part looks good to me. Ship it!

SammyK:
Fixing now :)

SammyK:
Made changes based on feedback. :)

I added a `FakeGraphApi/` folder to the tests where we can put all the mock Graph request/responses from here on. That'll keep our "fake Graph API" responses all in one place. Since this chunked upload feature requires at minimum 3 requests to Graph, I had to make the fake Graph API slightly more complex. Making it as its own class makes it much more understandable. :)

gfosco:
:+1: :+1: :+1: :+1:

codingtmd:
@gfosco , what's the release version for this change? I would like to make an announcement to our partners too.

gfosco:
@codingtmd discussing that here: https://github.com/facebook/facebook-php-sdk-v4/issues/504 It is not in a released version yet, should be 5.1.

PR #503, opened 2015-10-05
Don't convert a 0 timestamp, to support non-expiring tokens.

gfosco:
https://developers.facebook.com/bugs/917959338271504/


SammyK:
:+1:  Did you mean to submit two of these? #502 :)

PR #505, opened 2015-10-09
Update Graph version to v2.5

SammyK:
Graph [v2.5](https://developers.facebook.com/docs/apps/changelog) is out. :)


yguedidi:
For 6.0, we should stop updating the Graph version, it is a BC break :)
I think we should force the user to explicitly set a Graph version.

SammyK:
Yep! I've set up the [League OAuth 2 Facebook provider](https://github.com/thephpleague/oauth2-facebook] to throw an exception if the graph version isn't set.

We should start a 6.0 discussion soon. :)

PR #510, opened 2015-10-13
Complexity reduction

EmanueleMinotto:
To solve the `Facebook\Facebook` constructor complexity problem (defined on [Scrutinizer]()), I've splitted it into different factories:

* `Facebook\PseudoRandomString\PseudoRandomStringGeneratorFactory`
* `Facebook\HttpClients\HttpClientsFactory`
* `Facebook\PersistentData\PersistentDataFactory`

This should also allow you to move different detection logic into these factories and keeping some other classes' focus on their behaviour and responsabilities.

Another small change I did is the removal of an `\InvalidArgumentException` exception replicating a setter logic (afaik).

Of course I'll add related tests and documentation if the RFC is accepted.

### DIFF
--- a/src/Facebook/HttpClients/HttpClientsFactory.php
+++ b/src/Facebook/HttpClients/HttpClientsFactory.php
@@ -1,0 +1,98 @@
+<?php
+/**
+ * Copyright 2014 Facebook, Inc.
+ *
+ * You are hereby granted a non-exclusive, worldwide, royalty-free license to
+ * use, copy, modify, and distribute this software in source code or binary
+ * form for use in connection with the web services and APIs provided by
+ * Facebook.
+ *
+ * As with any software that integrates with the Facebook platform, your use
+ * of this software is subject to the Facebook Developer Principles and
+ * Policies [http://developers.facebook.com/policy/]. This copyright notice
+ * shall be included in all copies or substantial portions of the software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+ * THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ *
+ */
+namespace Facebook\HttpClients;
+
+use GuzzleHttp\Client;
+use InvalidArgumentException;
+
+class HttpClientsFactory
+{
+    private function __construct()
+    {
+        // a factory constructor should never be invoked
+    }
+
+    /**
+     * HTTP client generation.
+     *
+     * @param FacebookHttpClientInterface|Client|string|null $handler
+     *
+     * @throws Exception                If the cURL extension or the Guzzle client aren't available (if required).
+     * @throws InvalidArgumentException If the http client handler isn't "curl", "stream", "guzzle", or an instance of Facebook\HttpClients\FacebookHttpClientInterface.
+     *
+     * @return FacebookHttpClientInterface
+     */
+    public static function createHttpClient($handler)
+    {
+        if (!$handler) {
+            return self::detectDefaultClient();
+        }
+
+        if ($handler instanceof FacebookHttpClientInterface) {
+            return $handler;
+        }
+
+        if ('stream' === $handler) {
+            return new FacebookStreamHttpClient();
+        }
+        if ('curl' === $handler) {
+            if (!extension_loaded('curl')) {
+                throw new Exception('The cURL extension must be loaded in order to use the "curl" handler.');
+            }
+
+            return new FacebookCurlHttpClient();
+        }
+
+        if ('guzzle' === $handler && !class_exists('GuzzleHttp\Client')) {
+            throw new Exception('The Guzzle HTTP client must be included in order to use the "guzzle" handler.');
+        }
+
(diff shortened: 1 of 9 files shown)


yguedidi:
:+1: for reducing the complexity, :-1: on how it's done here.

I don't think we want those factories beind part of the SDK API.
I think you should move factrorie to private function of the `Facebook` master class.

EmanueleMinotto:
Imo depends on "how the final developers should use the library", because if the `Facebook\Facebook` must be the only starting point for them, than ok, but if the final developer can start from `Facebook\FacebookBatchRequest` (if they already know that they want just use the batch requests system) than the factories logic can be reused.

@yguedidi what do you think about this?

yguedidi:
I think that any guessing should go in the master class, the easy to use starting point for developers, and any internal thing (or call it "advanced usage") should be type hinted.

In the example of `Facebook\FacebookBatchRequest`, the advanced user should know he/she needs a `Facebook\FacebookClient`, that needs a `Facebook\HttpClients\FacebookHttpClientInterface`. Then he/she will find one or will implement one. :)

src/Facebook/HttpClients/HttpClientsFactory.php
```diff
@@ -0,0 +1,98 @@
...
+    {
+        if (!$handler) {
+            return self::detectDefaultClient();
+        }
+
+        if (is_object($handler) && $handler instanceof FacebookHttpClientInterface) {
+            return $handler;
+        }
+
+        if (is_string($handler) && 'stream' === $handler) {
```
stof: this condition can be simplified as ``'stream' === $handler``, with the same behavior, but a smaller cyclomatic complexity when computing it: you will only have 2 cases for this condition (the ``stream`` string vs anything else) instead of having 3 cases (the ``stream`` string vs any other string vs any other type)).

src/Facebook/HttpClients/HttpClientsFactory.php
```diff
@@ -0,0 +1,98 @@
...
+            }
+
+            return new FacebookCurlHttpClient();
+        }
+
+        if (is_string($handler) && 'guzzle' === $handler && !class_exists('GuzzleHttp\Client')) {
+            throw new Exception('The Guzzle HTTP client must be included in order to use the "guzzle" handler.');
+        }
+
+        if (is_object($handler) && $handler instanceof Client) {
```
stof: this can be simplified too. ``$handler instanceof Client`` handles non-object variables fine (they won't be considered instances)

src/Facebook/HttpClients/HttpClientsFactory.php
```diff
@@ -0,0 +1,98 @@
...
+            return new FacebookGuzzleHttpClient($handler);
+        }
+        if (is_string($handler) && 'guzzle' === $handler) {
+            return new FacebookGuzzleHttpClient();
+        }
+
+        throw new InvalidArgumentException('The http client handler must be set to "curl", "stream", "guzzle", or be an instance of Facebook\HttpClients\FacebookHttpClientInterface');
```
stof: the exception message does not actually describe all supported cases

tests/FacebookTest.php
```diff
@@ -177,7 +177,7 @@ public function testPersistentDataHandlerCanBeForced()
     }
 
     /**
-     * @expectedException \InvalidArgumentException
+     * @expectedException PHPUnit_Framework_Error
```
stof: this will be broken on PHP 7 because the behavior for failed typehints is not the same anymore (it throws an Error instead of triggering a catchable fatal error)

stof:
@yguedidi having separate factories makes sense at least for the guessing of the best implementation, as it makes the logic usable even when you use directly the lower-level API (instead of having to duplicate the logic).
And the current implementation makes it possible as ``HttpClientsFactory::create(null)``.
Btw, it may make sense to allow omitting the argument for such case if you decide that it is a meaningful use case. I would also change the name ``create``, because a static ``create`` method often creates an instance of the class itself, while it creates an HttpClient here, not an HttpClientsFactory.

tests/FacebookTest.php
SammyK: Just a clarification; failed argument parsing throws a [`TypeError`](http://php.net/typeerror) in PHP 7. [See example](https://3v4l.org/IYvad). :)

src/Facebook/HttpClients/HttpClientsFactory.php
SammyK: What is missing? I see all supported types described in the message.

SammyK:
I'm down with the factories. There was another guy I was showing the source code to at a recent PHP conference who suggested using factories as well.

> I would also change the name create, because a static create method often creates an instance of the class itself

@stof What convention are you referring to? On "php the right way" [the example uses a `create()` method in a factory](http://www.phptherightway.com/pages/Design-Patterns.html).

src/Facebook/HttpClients/HttpClientsFactory.php
EmanueleMinotto: I didn't include the `GuzzleHttp\Client` possibility, now included.

EmanueleMinotto:
@stof first of all: thank you for all those corrections, applied and now it should be ok.
I've also understood what you mean about the `create` method, but I have no idea for other names (considering the repeated pattern of *"automatic extraction of available tools"* an interface isn't so far, imo), do you have a suggestion for an alternative name?

gfosco:
You have my :+1:  :smiley:

stof:
@EmanueleMinotto I would use ``createHttpClient`` here

stof:
> a common interface isn't so far, imo

It is far. the only interface you could extract in your current implementation would be ``create(mixed $arg): mixed``, which does not make much sense.

EmanueleMinotto:
@stof you're right, now it's senseless an interface for them, thanks again.

@gfosco updated with tests and documentation, considering the internal usage I didn't touch `docs/` files, just added a line to the changelog, is it ok? I'm not sure about how to proceed for the documentation.

SammyK:
:+1: No need to worry about docs. All changes are internal and won't the public API unless we wanted to document the factories. But I think it might confuse newbies so if we decide to document it, it should probably be added to the reference only and not part of the "getting started" or anything.

Thanks for helping clean up the codebase @EmanueleMinotto!

EmanueleMinotto:
my pleasure @SammyK :)

gfosco:
Cool! :+1:

PR #514, opened 2015-10-15
Guzzle v6.1 compatibility

serhatdurum:
Guzzle v6.1 compatibility great job and many thanks for @SammyK

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,13 +12,13 @@
         }
     ],
     "require": {
-        "php": ">=5.4.0",
+        "php": ">=5.5.0",
         "ext-mbstring": "*"
     },
     "require-dev": {
         "phpunit/phpunit": "~4.0",
         "mockery/mockery": "~0.8",
-        "guzzlehttp/guzzle": "~5.0"
+        "guzzlehttp/guzzle": "~6.1"
     },
     "suggest": {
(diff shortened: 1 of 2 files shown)


gfosco:
Hmm.. Upping the minimum PHP version to 5.5.. thoughts @SammyK @yguedidi ?

gfosco:
Also this has failed Travis for all versions.

yguedidi:
I think this should be done in 6.0

SammyK:
Def a 6.0 thing. In fact I was hoping 6.0 could be Guzzle-only and we rip out all the custom HTTP client stuff. Now that the latest version of Guzzle is PSR-7, it shouldn't be changing much going forward (or so I hear from those with some insider information). :)

Dependency on Guzzle would make PHP SDK v6.0 dependent on composer which was the big drawback a few years ago. These days we could probably get away with it since even those without composer on their servers can still install everything locally and there's other workarounds we could discuss.

gfosco:
I'll close this for now... can be addressed when we're working towards 6.0.  Thanks!

SammyK:
Thanks for keeping on pushing for Guzzle 6 @serhatdurum! We'll definitely get it added eventually. :)

PR #516, opened 2015-10-16
Check curl extension

EmanueleMinotto:
The `FacebookCurlHttpClient` is using more than the `curl_init` function, so that class could be invoked without problems creating a fake `curl_init`: https://3v4l.org/CjcVS


SammyK:
:+1:

PR #517, opened 2015-10-16
Updating Composer install instructions

rdohms:
The `require` command is the best way to add new dependencies to a project and runs with as little side effects as possible.
Since last year it also binds properly to the latest version of the package. Its the recommended way to add new dependencies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,14 +10,10 @@ This repository contains the open source PHP SDK that allows you to access the F
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


SammyK:
I know that guy! Good to see you in the open-source world. :)

I'm :+1: for the change. You might want to consider updating [the docs](https://github.com/facebook/facebook-php-sdk-v4/blob/master/docs/sdk_getting_started.fbmd#L30-L47) as well.

rdohms:
done also, thanks for pointing it out

gfosco:
:+1:

PR #523, opened 2015-10-20
Fix parsing of batch headers

### LINKED ISSUE #521: FacebookResponse->getETag() returns null for batched response
I can get all the headers and find the ETag there, however the utility getETag() function does not work.

It looks like the headers value on a FacebookResponse is constructed differently depending if it is an unbatched or part of a batched request.

Dumping there headers I can see unbatched requests are an associative array eg-
['Header1Name' = 'Header1Value', ...]

Batched request headers are a list of key value entires eg-
[['name' = 'Header1Name', 'value' = Header1Value], ...]

I'm not sure if the appropriate fix would be to update the way the FacebookResponse is constructed so that the headers structure is consistent or to update the accessor methods to handle the difference in structures.

SammyK:
Fixes #521. :)

### DIFF
--- a/src/Facebook/FacebookBatchResponse.php
+++ b/src/Facebook/FacebookBatchResponse.php
@@ -102,7 +102,7 @@ class FacebookBatchResponse extends FacebookResponse implements IteratorAggregat
 
         $httpResponseBody = isset($response['body']) ? $response['body'] : null;
         $httpResponseCode = isset($response['code']) ? $response['code'] : null;
-        $httpResponseHeaders = isset($response['headers']) ? $response['headers'] : [];
+        $httpResponseHeaders = isset($response['headers']) ? $this->normalizeBatchHeaders($response['headers']) : [];
 
         $this->responses[$originalRequestName] = new FacebookResponse(
             $originalRequest,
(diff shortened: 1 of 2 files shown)


src/Facebook/FacebookBatchResponse.php
```diff
@@ -102,7 +102,7 @@ public function addResponse($key, $response)
 
         $httpResponseBody = isset($response['body']) ? $response['body'] : null;
         $httpResponseCode = isset($response['code']) ? $response['code'] : null;
-        $httpResponseHeaders = isset($response['headers']) ? $response['headers'] : [];
+        $httpResponseHeaders = isset($response['headers']) ? $this->normalizeBatchHeaders($response['headers']) : [];
```
yguedidi: useless private function, it just can be `array_column($response['headers'], 'value', 'name')` ;)

SammyK: You and your efficient ways. :) Since [`array_column()` is a PHP 5.5 feature](http://php.net/array_column), we'll need to wait to add this until PHP SDK v6.0 (which I'm assuming will require PHP 5.5.) :)

yguedidi: You're right!! My bad, I forgot that we support 5.4 in v5

yguedidi: About v6.0, it can even require at least 5.6 :D

SammyK: I like the way you think. :)

gfosco:
:+1: :+1:

PR #524, opened 2015-10-22
Removed duplicated logic (`createPseudoRandomStringGenerator`)

EmanueleMinotto:


SammyK:
:+1: 

If we embrace composer dependencies for v6, we can have all this CSPRNG stuff ripped out for the [PHP 7 CSPRNG polyfill](https://github.com/paragonie/random_compat). :)

EmanueleMinotto:
@SammyK it's up to you guys :) that polyfill seems really good!

PR #526, opened 2015-10-27
headers type + empty function

EmanueleMinotto:
* single internal type for response headers
* use the empty function instead of implicit casting


SammyK:
:+1:

PR #527, opened 2015-10-28
replaced deprecated method

EmanueleMinotto:
The method `makeGraphObject ` is marked as deprecated, so internally shouldn’t be used.

### DIFF
--- a/src/Facebook/GraphNodes/GraphObjectFactory.php
+++ b/src/Facebook/GraphNodes/GraphObjectFactory.php
@@ -68,7 +68,7 @@ class GraphObjectFactory extends GraphNodeFactory
      */
     public function makeGraphEvent()
     {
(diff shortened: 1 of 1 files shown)


src/Facebook/GraphNodes/GraphObjectFactory.php
```diff
@@ -62,13 +62,13 @@ public function makeGraphObject($subclassName = null)
     /**
      * Convenience method for creating a GraphEvent collection.
      *
-     * @return GraphEvent
+     * @return GraphNode
```
SammyK: This still returns a `GraphEvent` :)

EmanueleMinotto:
@SammyK fixed phpdoc, thanks :)

SammyK:
:+1:

PR #530, opened 2015-10-30
Fixed PHPDoc annotations

EmanueleMinotto:
`Facebook\Authentication\AccessToken` is accepted by both `Facebook\Authentication\OAuth2Client` and `Facebook\FileUpload\FacebookResumableUploader` because implements the `__toString` method, but it wasn't added in the PHPDoc annotations.


gfosco:
:+1:

PR #533, opened 2015-11-06
Update the GraphAlbum

devmsh:
generate the Node class based on the latest metadata from Graph API
metadata.

### DIFF
--- a/src/Facebook/GraphNodes/GraphAlbum.php
+++ b/src/Facebook/GraphNodes/GraphAlbum.php
@@ -35,12 +35,12 @@ class GraphAlbum extends GraphNode
      * @var array Maps object key names to Graph object types.
      */
     protected static $graphObjectMap = [
-        'from' => '\Facebook\GraphNodes\GraphUser',
+        'event' => '\Facebook\GraphNodes\GraphEvent',
         'place' => '\Facebook\GraphNodes\GraphPage',
     ];
 
     /**
-     * Returns the ID for the album.
+     * Returns the album ID.
      *
      * @return string|null
      */
@@ -60,7 +60,7 @@ class GraphAlbum extends GraphNode
     }
 
     /**
-     * Returns the number of photos in this album.
+     * Returns number of photos in this album.
      *
      * @return int|null
      */
@@ -70,9 +70,9 @@ class GraphAlbum extends GraphNode
     }
 
     /**
-     * Returns the ID of the album's cover photo.
+     * Returns album cover photo id.
      *
-     * @return string|null
+     * @return GraphNode|null
      */
     public function getCoverPhoto()
     {
@@ -90,16 +90,6 @@ class GraphAlbum extends GraphNode
     }
 
     /**
-     * Returns the time the album was updated.
-     *
-     * @return \DateTime|null
-     */
-    public function getUpdatedTime()
-    {
-        return $this->getField('updated_time');
-    }
-
-    /**
      * Returns the description of the album.
      *
      * @return string|null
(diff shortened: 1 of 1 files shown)


devmsh:
I will update the tests and docs once reviewed :)

gfosco:
Looks alright to me.. continue. :+1:

src/Facebook/GraphNodes/GraphAlbum.php
```diff
@@ -110,9 +100,9 @@ public function getDescription()
     }
 
     /**
-     * Returns profile that created the album.
+     * Returns the profile that created the album.
      *
-     * @return GraphUser|null
+     * @return GraphNode|null
```
SammyK: This should still be `GraphUser`: https://developers.facebook.com/docs/graph-api/reference/v2.5/album

src/Facebook/GraphNodes/GraphAlbum.php
```diff
@@ -35,12 +35,12 @@ class GraphAlbum extends GraphNode
      * @var array Maps object key names to Graph object types.
      */
     protected static $graphObjectMap = [
-        'from' => '\Facebook\GraphNodes\GraphUser',
```
SammyK: We need to keep `from`

src/Facebook/GraphNodes/GraphAlbum.php
devmsh: from the Graph API metadata
```
{
        "name": "from",
        "description": "The profile that created the album",
        "type": "user|page"
      },
```

The album may created by User or Page, so it must be returned as `GraphNode`, or as I suggest before, we can have something called `GraphProfile` as a super class of `GraphUser` and `GraphPage` to some common functionality like name, username, picture, ... etc.

So I remove the auto cast, and change the return type to `GraphNode` until we have another option?

src/Facebook/GraphNodes/GraphAlbum.php
devmsh: https://github.com/facebook/facebook-php-sdk-v4/pull/533#discussion_r44221751

src/Facebook/GraphNodes/GraphAlbum.php
SammyK: Gotcha. Well this would be a BC (see the failing test). So we probably need to figure a way to support these changes from Graph without breaking the SDK. @yguedidi Any thoughts?

yguedidi: What about something like:
```php
    /**
     * Returns the profile that created the album.
     *
     * @param bool @isPage true if you want a GraphPage instance
     *
     * @return GraphUser|GraphPage|null
     */
    public function getFrom($isPage = false)
    {
        if (!$isPage) {
             return $this->getField('from');
        }

        // return a casted to a GraphPage instance
    }
```

I'm not really sur. But for now, all users get a GraphUser instance, so I think we should keep this behavior as the default, and provide a way to get a GraphPage. :)

SammyK: :+1: to `@return GraphUser|GraphPage|null` docblock :)

gfosco:
Status on this, @devmsh ?

SammyK:
Hey @devmsh! Were you able to finish this one out? It'd be great to get this merged in for the next minor release. :)

SammyK:
Ping @devmsh. We cool with wrapping this one up? I'd like to merge it in without breaking the build on master. :)

devmsh:
Sorry, I was very busy recently.

I will make sure to update the PR maximum tomorrow, and I will review the current SDK status and resume my contributions :)

SammyK:
@devmsh - no worries at all! I've also be crazy busy. Thanks! :)

yguedidi:
@devmsh it's been a while... how are you? Do you think you could update this PR? :)

PR #535, opened 2015-11-10
[Bugfix] Facebook AccessToken is being validated correctly

lexakozakov:
That commit fixes the access token validation bug, when access token is not being validated at all.

### DIFF
--- a/src/Facebook/FacebookClient.php
+++ b/src/Facebook/FacebookClient.php
@@ -197,7 +197,7 @@ class FacebookClient
      */
     public function sendRequest(FacebookRequest $request)
     {
-        if (get_class($request) === 'FacebookRequest') {
+        if (get_class($request) === 'Facebook\FacebookRequest') {
             $request->validateAccessToken();
         }
 
--- a/tests/FacebookClientTest.php
+++ b/tests/FacebookClientTest.php
@@ -218,8 +218,8 @@ class FacebookClientTest extends \PHPUnit_Framework_TestCase
         $this->assertContains('multipart/form-data; boundary=', $headersSent['Content-Type']);
     }
 
-    public function testAFacebookRequestValidateAccessTokenIsNotProvided(){
-
+    public function testAFacebookRequestValidatesTheAccessTokenWhenOneIsNotProvided()
+    {
         $this->setExpectedException('Facebook\Exceptions\FacebookSDKException');
 
         $fbRequest = new FacebookRequest($this->fbApp, null, 'GET', '/foo');
--- a/tests/FacebookClientTest.php
+++ b/tests/FacebookClientTest.php
@@ -218,6 +218,14 @@ class FacebookClientTest extends \PHPUnit_Framework_TestCase
         $this->assertContains('multipart/form-data; boundary=', $headersSent['Content-Type']);
     }
 
+    public function testAFacebookRequestValidateAccessTokenIsNotProvided(){
+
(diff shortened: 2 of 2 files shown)


tests/FacebookClientTest.php
```diff
@@ -218,6 +218,14 @@ public function testARequestWithFilesWillBeMultipart()
         $this->assertContains('multipart/form-data; boundary=', $headersSent['Content-Type']);
     }
 
+    public function testAFacebookRequestValidateAccessTokenIsNotProvided(){
```
SammyK: `{` should be on next line.

Also method name should be `testAFacebookRequestValidatesTheAccessTokenWhenOneIsNotProvided()` :)

src/Facebook/FacebookClient.php
```diff
@@ -197,7 +197,7 @@ public function prepareRequestMessage(FacebookRequest $request)
      */
     public function sendRequest(FacebookRequest $request)
     {
-        if (get_class($request) === 'FacebookRequest') {
+        if (get_class($request) === 'Facebook\FacebookRequest') {
```
SammyK: It took me a while to remember why this `if` block is here. It's because the access token isn't required on the root request of a batch request. So this line might actually read better if it were changed to this:

```php
if (get_class($request) !== 'Facebook\FacebookBatchRequest') {
```

@yguedidi What are your thoughts?

lexakozakov: Hm, but $request could never be "FacebookBatchRequest". FacebookRequest does not implement any interface.

tests/FacebookClientTest.php
lexakozakov: Done, tnx.

src/Facebook/FacebookClient.php
yguedidi:
```php
if (!$request instanceof FacebookBatchRequest) {
```
;)
@lexakozakov FacebookBatchRequest extends FacebookRequest, so you can pass a FacebookBatchRequest to sendRequest

SammyK: There isn't an interface but `FacebookBatchRequest` extends from `FacebookRequest`. When you send a batch request it [eventually calls the `sendRequest()` method](https://github.com/facebook/facebook-php-sdk-v4/blob/master/src/Facebook/FacebookClient.php#L243-L246).

SammyK: @yguedidi You beat me to it again! :)

And yes to `instanceof`!

lexakozakov: Hm, "FacebookBatchRequest" improvement is not related to the fixed bug, isn't it?

SammyK: It is related since the `get_class()` implementation was broken, but this improvement also makes the code a lot more readable. Do you mind making the change to `instanceof`? :) This should be good to go after that.

lexakozakov: No, of course I don't mind to use instanceof. Why should I? :) 
I just wondering if I should make all changes related to "FacebookBatchRequest" in current pull request.

SammyK: Because this:

```php
if (!$request instanceof FacebookBatchRequest) {
```

Explains why access token validation gets skipped a lot better than this:

```php
if (get_class($request) === 'Facebook\FacebookRequest') {
```

It also prevents more hard-coded fully-qualified class names. :)

gfosco:
:+1:

PR #540, opened 2015-12-04
fix use of exception class

koenpunt:
Trying to initialize with `http_client_handler` set to `guzzle` without having Guzzle available (or `curl` without cURL) would result in a fatal error because there is no `Facebook\HttpClients\Exception` class:
```
PHP Fatal error:  Class 'Facebook\HttpClients\Exception' not found in ..../vendor/facebook/php-sdk-v4/src/Facebook/HttpClients/HttpClientsFactory.php on line 68
```

So I'd figure it should be a generic `Exception`.

### DIFF
--- a/src/Facebook/HttpClients/HttpClientsFactory.php
+++ b/src/Facebook/HttpClients/HttpClientsFactory.php
@@ -25,6 +25,7 @@ namespace Facebook\HttpClients;
 
 use GuzzleHttp\Client;
 use InvalidArgumentException;
+use Exception;
 
 class HttpClientsFactory
 {


EmanueleMinotto:
I forgot it, thanks! :)
Imo also a test should be added for every type of exception in that class.

SammyK:
:+1: Thanks!

I think the `Exception`'s in this class should really be `RuntimeException`'s, but that's for another PR. :)

koenpunt:
@EmanueleMinotto I believe tests are difficult (if not impossible) for `extension_loaded` and `class_exists`. Or the environment should be set up to not include the extension and/or class.

EmanueleMinotto:
@koenpunt that's true, but we could add a couple of optionally skipped tests if:

* for the test based on `extension_loaded`, the `@requires extension curl` annotation isn't satisfied
* for the test based on `class_exists`, the `class_exists` isn't satisfied in the test neither

What do you think about this? /cc @SammyK

gfosco:
If anyone wants to submit additional tests, I'm all for that.  Thanks. :+1:

PR #545, opened 2016-01-08
Added version number (v5) to headline

francimedia:
We received a report that people are confused by the repo name. `facebook-php-sdk-v4` indicates that the repo is for v4, not v5 or other versions.
To make clear that people can find v5 in this repo the latest version number should be part of the headline.


gfosco:
:+1:

Issue #546, opened 2016-01-15
Received a fatal because mb_substr was not found.


DragonBe:
Hi,

I received a fatal error because multibyte support was not enabled and because the application was looking for mb_string in namespace `Facebook\PseudoRandomString` it wasn't able to find it.

![screenshot 2016-01-15 14 48 22](https://cloud.githubusercontent.com/assets/282825/12355019/1f0d44fc-bb9a-11e5-8687-5d28c32f352e.png)

Aparently you only get the notification about the mbstring extension when you install through composer (which we didn't do because of our architecture).

```
composer install
Problem 1
    - The requested PHP extension ext-mbstring * is missing from your system. Install or enable PHP's mbstring extension.
```

Simple solution is to validate that the module is loaded in code and throw an exception if not (might break BC for everyone not using composer install. Even though this is a BC, I rather have an exception that's captured at the top-level than a **Fatal** killing all processes.

```
if(1 !== extension_loaded('mbstring')) {
    throw new \RuntimeException('Multibyte support required');
}
```

Then load the function with proper `\` root reference.

```
return \mb_substr(\bin2hex($binaryData), 0, $length);
```

SammyK:
Hey @DragonBe! Good to see you in the open source world. :)

I totally thought we killed the `mbstring` requirement a while ago since I thought we were only using it for the CSPRNG. And since the `bin2hex()` will always only generate `[0-9a-f]`, we don't need to worry about supporting > 256 bits.

But I did some searching and another place it's used is in `SignedRequest` to [validate the signed request signature](https://github.com/facebook/facebook-php-sdk-v4/blob/master/src/Facebook/SignedRequest.php#L269-L282) which also doesn't need > 256 bit support.

So in those two cases we could just change all references to `mb_substr()` to `substr()`.

The only place we *do* need `mbstring` is in the [cURL HTTP Client](https://github.com/facebook/facebook-php-sdk-v4/blob/master/src/Facebook/HttpClients/FacebookCurlHttpClient.php#L165-L173) *(ew)* but you can easily get around that by defining the `stream` HTTP client option when in instantiate the `\Facebook\Facebook` service to avoid using the cURL client.

```php
# Use the PHP stream implementation
$fb = new \Facebook\Facebook([
    'http_client_handler' => 'stream',
]);
```

[...]

PR #552, opened 2016-02-09
Remove mb-string dependency

SammyK:
Re #546

This kills the unnecessary mb-string dependency. In doing so I was able to rip out quite a bit of code & tests related to the overly-complicated cURL client implementation. While ripping out mb-string for signed request signature validation, I made a polyfill for the [`hash_equals()` function](http://php.net/manual/en/function.hash-equals.php) in PHP 5.6 *(which has been [vetted by infosec nerds](https://twitter.com/SammyK/status/697179032983482368))*. This allowed me to removed the duplicate code for comparing the CSRF token for redirect logins as well as signed request signature validations.

Also - the tests were failing out of the box when mcrypt, openssl or curl weren't installed. I fixed all that. :)

Make sure to `composer dump-autoload` before reviewing this one. :)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,8 +12,7 @@
         }
     ],
     "require": {
-        "php": ">=5.4.0",
-        "ext-mbstring": "*"
+        "php": ">=5.4.0"
     },
     "require-dev": {
         "phpunit/phpunit": "~4.0",
@@ -26,7 +25,8 @@
     "autoload": {
         "psr-4": {
             "Facebook\\": "src/Facebook/"
-        }
+        },
+        "files": ["src/Facebook/polyfills.php"]
     },
     "autoload-dev": {
         "psr-4": {
--- a/src/Facebook/Helpers/FacebookRedirectLoginHelper.php
+++ b/src/Facebook/Helpers/FacebookRedirectLoginHelper.php
@@ -235,27 +235,19 @@ class FacebookRedirectLoginHelper
     protected function validateCsrf()
     {
         $state = $this->getState();
+        if (!$state) {
+            throw new FacebookSDKException('Cross-site request forgery validation failed. Required GET param "state" missing.');
+        }
         $savedState = $this->persistentDataHandler->get('state');
-
-        if (!$state || !$savedState) {
-            throw new FacebookSDKException('Cross-site request forgery validation failed. Required param "state" missing.');
+        if (!$savedState) {
+            throw new FacebookSDKException('Cross-site request forgery validation failed. Required param "state" missing from persistent data.');
         }
 
-        $savedLen = strlen($savedState);
-        $givenLen = strlen($state);
-
-        if ($savedLen !== $givenLen) {
-            throw new FacebookSDKException('Cross-site request forgery validation failed. The "state" param from the URL and session do not match.');
+        if (\hash_equals($savedState, $state)) {
+            return;
         }
 
-        $result = 0;
-        for ($i = 0; $i < $savedLen; $i++) {
-            $result |= ord($state[$i]) ^ ord($savedState[$i]);
-        }
-
-        if ($result !== 0) {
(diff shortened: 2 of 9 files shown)


composer.json
```diff
@@ -12,8 +12,7 @@
         }
     ],
     "require": {
-        "php": ">=5.4.0",
-        "ext-mbstring": "*"
+        "php": ">=5.4.0"
```
EmanueleMinotto: Until PHP 7 is under the "allowed failures" section on Travis, could you use `"php": "^5.4"` please? (also the `>=` is unrecommended by Composer itself).

paragonie-scott:
While removing dependencies is a good thing, if the extension is available, it would be *wise* to use `mb_strlen($str, '8bit');` instead of `strlen()` because of [function overloading](https://secure.php.net/manual/en/mbstring.overload.php).

```php

    function sammy_hash_equals($knownString, $userString)
    {
        $kLen = strlen($knownString);
        $uLen = strlen($userString);
        if ($kLen !== $uLen) {
            return false;
        }
        $result = 0;
        for ($i = 0; $i < $kLen; $i++) {
            $result |= (ord($knownString[$i]) ^ ord($userString[$i]));
        }
        // They are only identical strings if $result is exactly 0...
        return 0 === $result;
    }
// 8 chars but 32 bytes
$hashA = "\xF0\x9D\x92\xB3" . "\xF0\x9D\xA5\xB3" .
         "\xF0\x9D\x92\xB3" . "\xF0\x9D\xA5\xB3" .
         "\xF0\x9D\x92\xB3" . "\xF0\x9D\xA5\xB3" .
         "\xF0\x9D\x92\xB3" . "\xF0\x9D\xA5\xB3";

$hashB = "\xF0\x9D\x92\xB3" . "\xF0\x9D\xA5\xB3" .
         "\xF0\xAD\x9F\xC0" . "\xF0\xAD\x9F\xC0" .
         "\xF0\xAD\x9F\xC0" . "\xF0\xAD\x9F\xC0" .
         "\xF0\xAD\x9F\xC0" . "\xF0\xAD\x9F\xC0";

var_dump(sammy_hash_equals($hashA, $hashB));
```

If you run this unit test with `mbstring.func_overload` set to `2`, `3,` or `7`, this will return `true`. This is useful if an attacker wants to guess a MAC to forge an arbitrary message. (Since this is a global polyfill function, it's not inconceivable that someone would attempt to use this to validate a HMAC in an encryption library in the same application. Hello, chosen-ciphertext attacks.)

The attack strategy requires a little bit of luck (all bytes in the string must coincide with eight 4-byte UTF-8 sequence + the first 8 bytes must be identical). However, this reduces the cost of a birthday attack from ~2^128 (against HMAC-SHA-256) to about 2^65 (assuming 11 bits needed per 4-byte sequence, but the first 8 need to all match, and 2^(n/2) is the 50% mark for birthday collisions).

https://en.wikipedia.org/wiki/UTF-8#Description

composer.json
SammyK: Good point! Let's fix this issue in a separate PR. :)

SammyK:
Thanks @paragonie-scott! Pulled your changes in. :)

composer.json
EmanueleMinotto: Done in https://github.com/facebook/facebook-php-sdk-v4/pull/554 :)

yguedidi:
@SammyK looks good to me!
I'll just suggest to use the new symfony/polyfill repository for PHP polyfills instead of rewriting them. You can require the subpackage `"symfony/polyfill-php56": "^v1.1"` that provide a polyfill for `hash_equals` :)

SammyK:
@yguedidi Yeah I'd rather pull it in via composer as well. But since historically @gfosco has been against composer dependancies, let's try to butter him up for allowing them in v6. :D

SammyK:
Ping @gfosco. This one is ready to merge in when you get a sec. :)

PR #556, opened 2016-02-11
Add better error checking for app IDs as int when greater than PHP_INT_MAX

### LINKED ISSUE #532: Broken for large app Id values
The SDK is broken in a few places for large App Ids that get formatted in exponential notation by PHP.

I've patched my version locally with the following fixes:

OAuth2Client.php line 138 and 288:
            'client_id' => sprintf("%.0f", $this->app->getId()),

FacebookSignedRequestFromInputHelper.php line 160:

        $cookieName = sprintf('fbsr_%.0f', $this->app->getId());
        
        if (isset($_COOKIE[$cookieName])) {
            return $_COOKIE[$cookieName];
        }

Hopefully someone will pick this up and incorporate it in the code.

SammyK:
Fixes #532 

Facebook app ID's are really big and on 32-bit systems many are bigger than `PHP_INT_MAX` causing integer overflow issues. We can't fix when people overflow an int:

```php
$appId = PHP_INT_MAX + 10;
```

But we can at least give them better error messages and prevent overflows when `FacebookApp` is unserialized. :)

### DIFF
--- a/src/Facebook/FacebookApp.php
+++ b/src/Facebook/FacebookApp.php
@@ -24,6 +24,7 @@
 namespace Facebook;
 
 use Facebook\Authentication\AccessToken;
+use Facebook\Exceptions\FacebookSDKException;
 
 class FacebookApp implements \Serializable
 {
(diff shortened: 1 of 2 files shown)


SammyK:
@gfosco This bug fix is also ready to go. :)

PR #558, opened 2016-02-11
Making sure that AccessToken::__toString returns string

oleg-andreyev:
To avoid issue like: "Catchable Fatal Error: Method Facebook\Entities\AccessToken::__toString() must return a string value"

### DIFF
--- a/src/Facebook/Entities/AccessToken.php
+++ b/src/Facebook/Entities/AccessToken.php
@@ -65,7 +65,7 @@ class AccessToken
    */
   public function __construct($accessToken, $expiresAt = 0, $machineId = null)
   {
-    $this->accessToken = $accessToken;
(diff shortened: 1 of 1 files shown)


SammyK:
:+1: This could be updated for v5 as well. :)

yguedidi:
I'm :-1:, you duplicate the PHPDoc. Do you cast all parameters to the expected type? :)
Don't respect the PHPDoc is a developer responsability, he/she violating our API contract.

SammyK:
Oh the woes of a ducked-typed language. :)

SammyK:
Thanks for the contribution @oleg-andreyev! I'm going to go with @yguedidi on this one and close this one out. If you have any other ideas or PR's in your pocket, feel free to send another PR! :)

PR #560, opened 2016-02-17
Update get method to support parameters

phwd:
Given that this SDK already supports parameters in POST and
DELETE requests, it seems, the consistent action will be to include
parameters in the GET request.

The only caveat here (which also applies for POST and DELETE) being
when the access token is explicitly supplied, the params array must
also be supplied (even as an empty array)

e.g. `$response = $fb->delete('/{post-id}', '{explicit_access_token}');`
The above will fail asking for the array since the second input
is the params array. So for the above to work it's actually

e.g. `$response = $fb->delete('/{post-id}', [], '{explicit_access_token}');`

Where [] is an empty array or can be supplied params such as user when
unblocking a page as mentioned in #379

The docs have been updated to mention this.

### DIFF
--- a/docs/Facebook.fbmd
+++ b/docs/Facebook.fbmd
@@ -31,12 +31,12 @@ $response = $fb->post('/me/feed', ['message' => 'Foo message']);
 $response = $fb->delete('/{node-id}');
 ~~~~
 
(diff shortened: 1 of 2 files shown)


SammyK:
Thanks for the PR! This would be something that would have to be slated for v6 of the PHP SDK as it is a breaking change.

SammyK:
Thanks again for this PR @phwd! Since it's a breaking change to the public API, I have to close this one out for v5, but this feature will certainly be on the list of v6 since it has been requested so often. :)

PR #571, opened 2016-04-12
Bump Graph version to v2.6

SammyK:
[Graph v2.6](https://developers.facebook.com/docs/apps/changelog) is the new hotness. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ Simple GET example of a user's profile.
 $fb = new Facebook\Facebook([
   'app_id' => '{app-id}',
   'app_secret' => '{app-secret}',
-  'default_graph_version' => 'v2.5',
+  'default_graph_version' => 'v2.6',
   //'default_access_token' => '{access-token}', // optional
 ]);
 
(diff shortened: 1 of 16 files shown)


yguedidi:
@SammyK you know this is a BC break in fact? ;)

SammyK:
Yes, for those who are using the default fall-back version of Graph. That's why I wanted the next version of the SDK to throw an exception if you don't provide a default Graph version. But in the docs all the examples have the Graph version in the options array so for the copy-pasters, they'll have their Graph version set.

So yeah, it's BC for a very small subset, but I feel like we decided to keep bumping the Graph version every time a new one came out. But I'm cool with just keeping v2.5 for the default fallback in `Facebook` and just updating the docs to v2.6. Then the developers who are using the fall-back option will get upgraded to the next version of Graph when v2.5 is force-upgraded on April 12, 2018.

What do you guys think?

yguedidi:
> That's why I wanted the next version of the SDK to throw an exception if you don't provide a default Graph version.

Me too!

> I feel like we decided to keep bumping the Graph version every time a new one came out.

You're right, I'm not really against this change as we used to do it in the 5.x major version. I said it more like a reminder for 6.x

gfosco:
@SammyK @yguedidi You're now core collaborators with write privileges.  Wield your powers responsibly and for great justice.  🎉

SammyK:
You the man @gfosco! :)

PR #574, opened 2016-04-19
Updating logic for x_forwarded_host

gfosco:
@SammyK @yguedidi If you can review/comment/improve that would be great.

We got a report that x_forwarded_host could be used maliciously (in non-reverse-proxy situations)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
 vendor/
 composer.lock
 tests/FacebookTestCredentials.php
-
--- a/src/Facebook/Url/FacebookUrlDetectionHandler.php
+++ b/src/Facebook/Url/FacebookUrlDetectionHandler.php
@@ -95,8 +95,8 @@ class FacebookUrlDetectionHandler implements UrlDetectionInterface
     protected function getHostName()
     {
         // Check for proxy first
-        if ($this->isValidForwardedHost() && $host = $this->getHeader('X_FORWARDED_HOST')) {
-            $elements = explode(',', $host);
+        if ($header = $this->getHeader('X_FORWARDED_HOST') && $this->isValidForwardedHost($header)) {
+            $elements = explode(',', $header);
             $host = $elements[count($elements) - 1];
         } elseif (!$host = $this->getHeader('HOST')) {
             if (!$host = $this->getServerVar('SERVER_NAME')) {
@@ -164,20 +164,18 @@ class FacebookUrlDetectionHandler implements UrlDetectionInterface
     /**
      * Checks if the value in X_FORWARDED_HOST is a valid hostname
      * Could prevent unintended redirections
+     *
+     * @param string $header
+     *
+     * @return boolean
      */
-    protected function isValidForwardedHost()
+    protected function isValidForwardedHost($header)
     {
-        $host = $this->getHeader('X_FORWARDED_HOST');
-        if (!$host) {
-          return false;
-        }
-
-        $elements = explode(',', $host);
+        $elements = explode(',', $header);
         $host = $elements[count($elements) - 1];
         
-        return return (preg_match("/^([a-z\d](-*[a-z\d])*)(\.([a-z\d](-*[a-z\d])*))*$/i", $domain_name) //valid chars check
-            && preg_match("/^.{1,253}$/", $domain_name) //overall length check
-            && preg_match("/^[^\.]{1,63}(\.[^\.]{1,63})*$/", $domain_name) ); //length of each label
+        return preg_match("/^([a-z\d](-*[a-z\d])*)(\.([a-z\d](-*[a-z\d])*))*$/i", $host) //valid chars check
+            && 0 < strlen($host) && strlen($host) < 254 //overall length check
+            && preg_match("/^[^\.]{1,63}(\.[^\.]{1,63})*$/", $host); //length of each label
     }
-
 }
--- a/src/Facebook/Url/FacebookUrlDetectionHandler.php
+++ b/src/Facebook/Url/FacebookUrlDetectionHandler.php
@@ -95,7 +95,7 @@ class FacebookUrlDetectionHandler implements UrlDetectionInterface
     protected function getHostName()
     {
         // Check for proxy first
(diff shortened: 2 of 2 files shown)


src/Facebook/Url/FacebookUrlDetectionHandler.php
```diff
@@ -95,7 +95,7 @@ protected function protocolWithActiveSsl($protocol)
     protected function getHostName()
     {
         // Check for proxy first
-        if ($host = $this->getHeader('X_FORWARDED_HOST')) {
+        if ($this->isValidForwardedHost() && $host = $this->getHeader('X_FORWARDED_HOST')) {
```
yguedidi: I'd write `if ($header = $this->getHeader('X_FORWARDED_HOST') && $this->isValidForwardedHost($header))`. It will also simplify the function

src/Facebook/Url/FacebookUrlDetectionHandler.php
```diff
@@ -160,4 +160,24 @@ protected function getHeader($key)
...
+
+    /**
+     * Checks if the value in X_FORWARDED_HOST is a valid hostname
+     * Could prevent unintended redirections
+     */
+    protected function isValidForwardedHost()
+    {
+        $host = $this->getHeader('X_FORWARDED_HOST');
+        if (!$host) {
+          return false;
```
yguedidi: BTW, wrong indentation, but this will be removed when applying my above comment

src/Facebook/Url/FacebookUrlDetectionHandler.php
```diff
@@ -160,4 +160,24 @@ protected function getHeader($key)
...
+        }
+
+        $elements = explode(',', $host);
+        $host = $elements[count($elements) - 1];
+        
+        return return (preg_match("/^([a-z\d](-*[a-z\d])*)(\.([a-z\d](-*[a-z\d])*))*$/i", $domain_name) //valid chars check
```
yguedidi: must be `$host` instead of `$domain_name`, same below

SammyK: Double `return` keyword is making the build fail. :)

src/Facebook/Url/FacebookUrlDetectionHandler.php
```diff
@@ -160,4 +160,24 @@ protected function getHeader($key)
...
+            && preg_match("/^.{1,253}$/", $domain_name) //overall length check
```
yguedidi: `&& 0 < strlen($host) && strlen($host) < 254` looks clearer to me, and avoid the cost of a `preg_match` for length tests

src/Facebook/Url/FacebookUrlDetectionHandler.php
yguedidi: one `return` must be removed. BTW, parenthesis arround all conditions can be removed

yguedidi:
@gfosco from where you get the regexs?

src/Facebook/Url/FacebookUrlDetectionHandler.php
```diff
@@ -160,4 +160,24 @@ protected function getHeader($key)
...
+            && preg_match("/^[^\.]{1,63}(\.[^\.]{1,63})*$/", $domain_name) ); //length of each label
+    }
+
```
yguedidi: extra line :)

gfosco:
Ok, will do a more thorough job... I just quickly jammed this together using VIM, as I have a new machine just today and haven't set everything up.   and I haven't done PHP in a while...

.gitignore
```diff
@@ -1,4 +1,4 @@
 vendor/
 composer.lock
 tests/FacebookTestCredentials.php
-
+.idea/
```
yguedidi: must be removed, this should be part of a [global gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore), it's not related to the project

src/Facebook/Url/FacebookUrlDetectionHandler.php
```diff
@@ -160,4 +160,23 @@ protected function getHeader($key)
...
+     *
+     * @return boolean
+     */
+    protected function isValidForwardedHost($header)
+    {
+        $elements = explode(',', $header);
+        $host = $elements[count($elements) - 1];
+        
+        return preg_match("/^([a-z\d](-*[a-z\d])*)(\.([a-z\d](-*[a-z\d])*))*$/i", $host) //valid chars check
+            && preg_match("/^.{1,253}$/", $host) //overall length check
```
yguedidi: looks like you forgot this comment: `&& 0 < strlen($host) && strlen($host) < 254` looks clearer to me, and avoid the cost of a `preg_match` for length tests

src/Facebook/Url/FacebookUrlDetectionHandler.php
```diff
@@ -160,4 +160,23 @@ protected function getHeader($key)
...
+     * @param string $header
+     *
+     * @return boolean
+     */
+    protected function isValidForwardedHost($header)
+    {
+        $elements = explode(',', $header);
+        $host = $elements[count($elements) - 1];
+        
+        return preg_match("/^([a-z\d](-*[a-z\d])*)(\.([a-z\d](-*[a-z\d])*))*$/i", $host) //valid chars check
```
tgallice: Hi I miss maybe something, but this regex can not be replaced by `preg_match("/^[a-z\d\.-]+$/i", $host)` ?

yguedidi: @tgallice No it can't, your regex will allow a host to start with `.` or `-`.
BTW I think it can be replace with something like `/^[a-z\d][a-z\d-]*(?:\.[a-z\d][a-z\d-]*)*$/i`, what do you think about it? ping @gfosco @SammyK

stof: the regex is not equivalent either. The current regex does not allow to have a dash followed by a dot, while your regex allows it @yguedidi

PR #582, opened 2016-05-09
Avoids using unserialize to prevent Object Injection security issue.

diegoquinteiro:
Rewrites the serialization methods to use json_encode / json_decode to prevent Object Injection attacks.

See more: https://www.owasp.org/index.php/PHP_Object_Injection

This is also a request from WordPress to allow the SDK to be used in plugins on WordPress.com and WordPress VIP:

https://vip.wordpress.com/documentation/code-review-what-we-look-for/#serializing-data

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,10 @@ Starting with version 5, the Facebook PHP SDK follows [SemVer](http://semver.org
 ## 5.0.x
 
 Version 5 of the Facebook PHP SDK is a complete refactor of version 4. It comes loaded with lots of new features and a friendlier API.
-
+- 5.1.4
+  - Breaking changes
+    - Changes the serialization method of FacebookApp
+      - FacebookApps serialized by versions prior 5.1.4 cannot be unserialized by this version
 - 5.0 (2015-??-??)
   - New features
     - Added the `Facebook\Facebook` super service for an easier API
@@ -22,8 +25,8 @@ Version 5 of the Facebook PHP SDK is a complete refactor of version 4. It comes 
       - Many improvements to the Graph node subtypes
     - New injectable interfaces
       - Added a `PersistentDataInterface` for custom persistent data handling
-      - Added a `PseudoRandomStringGeneratorInterface` for customizable CSPRNG's 
-      - Added a `UrlDetectionInterface` for custom URL-detection logic 
+      - Added a `PseudoRandomStringGeneratorInterface` for customizable CSPRNG's
+      - Added a `UrlDetectionInterface` for custom URL-detection logic
   - Codebase changes
     - Moved exception classes to `Exception\*` directory
     - Moved response collection objects to `GraphNodes\*` directory
--- a/src/Facebook/FacebookApp.php
+++ b/src/Facebook/FacebookApp.php
@@ -84,7 +84,7 @@ class FacebookApp implements \Serializable
      */
     public function serialize()
     {
-        return serialize([$this->id, $this->secret]);
+        return json_encode([$this->id, $this->secret]);
     }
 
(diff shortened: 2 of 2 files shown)


SammyK:
Thanks @diegoquinteiro! This would be a breaking change but a BC for security is OK in my book. I'd like to hear @yguedidi and @sarciszewski's thoughts on this.

paragonie-scott:
Sadly, JSON alone isn't a great alternative either. The HashDoS still hasn't been mitigated. (It's a much lower concern than RCE, of course!)

Questions:

* Do users ever get control over `$this->id` or `$this->secret`?
* Are there any examples of people using this SDK in a vulnerable manner? (i.e. passing user input into the `unserialize()` method?)

paragonie-scott:
Given that this is being used just to store two properties, and the output of `unserialize()` is being assigned to two variables via `list()`, we could swap it out for `explode()` and `implode()`.

This side-steps a [whole lot of problems](https://paragonie.com/blog/2016/04/securely-implementing-de-serialization-in-php).

SammyK:
I love the `explode()` and `implode()` idea!

To answer you questions:

- Users can set the app ID and secret via the constructor.
- I don't know of any specific instances of users using the SDK in a vulnerable manner like this.

paragonie-scott:
What characters are allowed in both? If they're both limited, we can make Mr. Torgue proud.

diegoquinteiro:
Thank you guys for the feedback. I went head and changed it to use implode/explode approach.

paragonie-scott:
As long as `|` isn't a valid character for the ID or Secret, LGTM. :+1:

diegoquinteiro:
@SammyK why is this a breaking change? As far as I understand, the user is not supposed to use the output of `serialize` in any way other then using it as a parameter for `unserialize`. Is that accurate?

@paragonie-scott `|` is already used as a separator for these same fields on https://github.com/facebook/facebook-php-sdk-v4/blob/fix-object-injection/src/Facebook/FacebookApp.php#L77

paragonie-scott:
Good call. :)

diegoquinteiro:
@SammyK nevermind, I figured this is breaking as objects serialized in the current version will not be unserializeble after this change.

What about merging it and releasing a new version?

SammyK:
@diegoquinteiro I'm actually in favor of pushing the BC through in a `PATCH` release as it's security related and therefore important. And I can't imagine that many people are actually serializing the `FacebookApp` entity anyways.

diegoquinteiro:
@SammyK agreed. Should I merge it myself? Also, who's is usually tagging/generating the version?

SammyK:
@diegoquinteiro Once we get a 👍 from @yguedidi I can merge it in. :)

yguedidi:
I'm OK with the security patch and make it part of a patch release, but I'm not sure about the way to do it...

As we must introduce a BC break (for security reasons), why not just drop the ability to serialize the `FacebookApp` entity?
I can't remember why we make it serializable, and it seems to not be a use case...

BTW, as it's a BC break in a patch release, you must update the changelog with a visible annoucement! :)

diegoquinteiro:
@yguedidi I've updated the changelog. I did not remove the methods altogether as keeping them (while still changing the implementation) potentially breaks less cases then completely removing it.

I know there is probably few to no people using the methods, but as this is a PATCH I'd try to minimize the impact as much as possible.

What do you think? Can we merge it?

diegoquinteiro:
@SammyK can you merge/release? @yguedidi gave his 👍  to the patch =)

gfosco:
Merging and tagging a release for this.  Thanks @diegoquinteiro 👍

PR #589, opened 2016-05-21
Fix: Reduce visibility of setUp()

localheinz:
This PR

* [x] reduces the visibility of `setUp()` from `public` to `protected`

🙎 This is the visibility as declared on the parent; there's no need to increase it.


SammyK:
Thanks! :)

PR #592, opened 2016-05-21
Fix: Remove unused local variables

localheinz:
This PR

* [x] removes unused local variables in tests

💁 The variables are never actually assigned a value as during the evaluation of the expressions to the right, exceptions are thrown, so they can safely be removed, right?


SammyK:
Indeed - thanks! :)

PR #593, opened 2016-05-21
Fix: Attempt to cache dependencies between builds

localheinz:
This PR

* [x] attempts to cache dependencies installed with `composer` between builds

💁 Ain't nobody got time for that, right?!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,12 +9,16 @@ php:
 
 sudo: false
 
+cache:
+  directories:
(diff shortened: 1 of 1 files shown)


SammyK:
Ha! You crack me up. I don't have much experience with this one so I'd like to get a 👍 from @yguedidi before merging. :)

yguedidi:
Looks good to me :+1:

PR #594, opened 2016-05-21
Fix: Require test code to also be PSR-2-compliant

localheinz:
This PR

* [x] extracts a configuration file for `phpcs`
* [x] requires code in `tests/` to also be PSR-2 compliant 
* [x] runs `phpcbf` and extracts each class definition into its own file
* [x] moves extracted test assets (as well as existing test assets) into `Fixtures` namespace

💁 There really isn't any plausible reason for applying different standards for production and test code, is there?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,5 +21,5 @@ install:
   - travis_retry composer install --prefer-dist --no-interaction
 
 script:
-  - vendor/bin/phpcs src --standard=psr2 -spn
+  - vendor/bin/phpcs
   - vendor/bin/phpunit --coverage-text --exclude-group integration
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -1,6 +1,7 @@
 <?xml version="1.0"?>
 <ruleset>
     <file>src/</file>
+    <file>tests/</file>
     <arg value="spn" />
     <rule ref="PSR2" />
 </ruleset>
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -1,0 +1,6 @@
+<?xml version="1.0"?>
+<ruleset>
+    <file>src/</file>
+    <arg value="spn" />
+    <rule ref="PSR2" />
+</ruleset>
--- a/tests/PseudoRandomString/MyFooBarPseudoRandomStringGenerator.php
+++ b/tests/PseudoRandomString/MyFooBarPseudoRandomStringGenerator.php
@@ -23,23 +23,9 @@
  */
 namespace Facebook\Tests\PseudoRandomString;
 
-class PseudoRandomStringGeneratorTraitTest extends \PHPUnit_Framework_TestCase
-{
-    /**
-     * @expectedException \InvalidArgumentException
-     */
-    public function testAnInvalidLengthWillThrow()
-    {
-        $prsg = new MyFooBarPseudoRandomStringGenerator();
-        $prsg->validateLength('foo_len');
-    }
+use Facebook\PseudoRandomString\PseudoRandomStringGeneratorTrait;
 
-    /**
-     * @expectedException \InvalidArgumentException
-     */
-    public function testALengthThatIsNotAtLeastOneCharacterWillThrow()
-    {
-        $prsg = new MyFooBarPseudoRandomStringGenerator();
-        $prsg->validateLength(0);
-    }
+class MyFooBarPseudoRandomStringGenerator
+{
+    use PseudoRandomStringGeneratorTrait;
 }
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -1,6 +1,7 @@
 filter:
     paths:
         - 'src/*'
+        - 'tests/*'
 
 tools:
     php_code_sniffer:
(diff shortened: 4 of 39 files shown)


SammyK:
This is great - thanks! Since it's a bigger one, I'll let @yguedidi give it a once over as well before merging. :)

yguedidi:
* OK to extract phpcs config, but in a `phpcs.xml.dist` file as it's the default loaded one, and can be overided by a git ignored `phpcs.xml` one
* Not sure about using PSR-2 in tests...
  * First they're removed in dist zip
  * Second it's allow us to have multiple classes in one file for small fixture classes instead of many files
  * Finally, look at PhpSpec, a great API design tool, tests are written like `public function it_do_this_thing()` which is not PSR-2 compliant, then how to use it?

localheinz:
@yguedidi 

> OK to extract phpcs config, but in a phpcs.xml.dist file as it's the default loaded one, and can be overided by a git ignored phpcs.xml one

Addressed.

> Second it's allow us to have multiple classes in one file for small fixture classes instead of many files.

Can't really see how that is an advantage. Can you explain?

> Finally, look at PhpSpec, a great API design tool, tests are written like `public function it_do_this_thing()` which is not PSR-2 compliant, then how to use it?

These are the options: 

1. don't use `phpspec` (`phpunit` works just fine) and require test code to be PSR-2-compliant
2. use `phpspec`, and don't require test code to be PSR-2-compliant
3. don't require test code to be PSR-2-compliant

In regard to 2.: We're not using it here, so it's not a problem.
In regard to 3.: Certainly an option, but not a good one - consistency wins.

.travis.yml
```diff
@@ -21,5 +21,5 @@ install:
   - travis_retry composer install --prefer-dist --no-interaction
 
 script:
-  - vendor/bin/phpcs src --standard=psr2 -spn
+  - vendor/bin/phpcs --standard=phpcs.xml.dist
```
yguedidi: can just be `vendor/bin/phpcs`

phpcs.xml.dist
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0"?>
+<ruleset>
+    <file>src/</file>
+    <file>tests/</file>
+    <arg value="spn"/>
```
yguedidi: missinf space before closing tag

yguedidi:
@localheinz I'm not really against using PSR-2 in our current test files, I just think it's not worth it.
BTW, If you really want to separate fixture classes in their own file, please put them under a `Fixtures\` namespace.

localheinz:
@yguedidi 

Updated!

tests/GraphNodes/Fixtures/MyFooGraphNode.php
```diff
@@ -0,0 +1,33 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+ * THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ *
+ */
+namespace Facebook\Tests\GraphNodes\Fixtures;
```
yguedidi: `Fixtures\` must be the top level namespace, so this must be `Facebook\Tests\Fixtures\GraphNodes`. Same for other files. It's helpful to put all fixtures together :)

localheinz:
@yguedidi 

Amended!

tests/PseudoRandomString/MyFooBarPseudoRandomStringGenerator.php
```diff
@@ -0,0 +1,31 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+ * THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ *
+ */
+namespace Facebook\Tests\PseudoRandomString;
```
yguedidi: @localheinz You missed this one :)

localheinz:
@yguedidi 

And amended! 😄

yguedidi:
Looks good :)

SammyK:
Boom! Thanks guys! :)

localheinz:
Thank you, @SammyK and @yguedidi!

PR #597, opened 2016-05-25
Corrected phpDoc comment for binToHex

NinoSkopac:
The method doesn't actually throw \InvalidArgumentException


yguedidi:
Looks good

SammyK:
Thanks! :)

PR #601, opened 2016-06-01
Always end if/elseif with an else

NinoSkopac:


yguedidi:
:-1: You don't allow any more to use the current dir as a default base dir, it forces us to define the constant

SammyK:
Thanks for the contribution @NinoSkopac! But I'm also 👎 for this.

PR #602, opened 2016-06-01
Avoid "Cannot redeclare class"

NinoSkopac:
I am not very knowledgeable of this SDK, so this may be a stupid idea (tests will show), but if you opt in for require_once instead of require, you can't have the same class file re-required, which is a good thing.

### DIFF
--- a/src/Facebook/autoload.php
+++ b/src/Facebook/autoload.php
@@ -74,6 +74,6 @@ spl_autoload_register(function ($class) {
 
     // if the file exists, require it
     if (file_exists($file)) {
-        require $file;
+        require_once $file;
     }
 });


yguedidi:
:-1: This is useless, as the autoloader is called for a class that is not loaded yet by PHP, once it's loaded, our autoload callback is no more called for that class :)

SammyK:
Hey @NinoSkopac! Thanks for another PR! But again, I'm with @yguedidi on this one. The autoloader is based off of an [official example](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-4-autoloader-examples.md#closure-example) from the PHP-FIG. In addition to [`require_once` being slower than `require`](http://stackoverflow.com/questions/186338/why-is-require-once-so-bad-to-use), the autoloader won't be invoked again once a class is loaded so it's redundant to have `require_once`. :)

NinoSkopac:
Yep, my bad <3

PR #612, opened 2016-06-11
Update README.md - PHP version >= 5.6

antarasi:
In src/Facebook/SignedRequest.php:271 hash_equals function is used, which was added to PHP 5 >= 5.6.0
http://php.net/manual/en/function.hash-equals.php


SammyK:
Hey @antarasi! The SDK includes a [polyfill for that](https://github.com/facebook/facebook-php-sdk-v4/blob/master/src/Facebook/polyfills.php). :)

Issue #583, opened 2016-05-09
Reopen:  Getting different values for $_SESSION["FBRLH_state"] and $_GET['state'] on fb callback


spedinfargo:
Hi,

This elusive error is still happening for numerous people (myself included):  **Facebook SDK returned an error: Cross-site request forgery validation failed. The "state" param from the URL and session do not match.**

As mentioned in #529, there is something very strange happening with the FBRLH_state session variable.  As soon as it gets back to my callback script, its value has been overwritten by a new random value.

I've tried to narrow it down a bit - maybe some more notes can help?

I have sites hosted on two providers:  one works with all browsers that I've tested, one is broken in these scenarios:

- Windows (broken):  Chrome, Firefox
- Windows (OK):  Internet Explorer, Edge
- Android (broken):  Chrome, Maxthon
- Android (OK):  Opera
- iOS (OK):  Safari, Chrome
- OS X (broken):  Safari

The site that works is running PHP 5.5.27 and I manually downloaded the Facebook SDK (which looks like it is still 5.0.0).  Unfortunately it looks like phpinfo() is disabled there and I'm betting I don't have access to the php.ini.

On the broken environment, I was originally using PHP 5.6 with the Facebook SDK installed using Composer, which I believe is the most current (5.1.2?).  However, I switched to PHP 5.5 using CPanel (confirmed the change) and also manually installed the 5.0.0 version (same as the site that works).  In other words, without changing any php.ini settings, I was hoping to create a similar environment.

On the site that doesn't work, I created a pretty standard Login page:

```
<?php
require_once __DIR__ . '/vendor/autoload.php';

session_start();

$fb = new Facebook\Facebook([
  'app_id' => '999999999', // Replace {app-id} with your app id
  'app_secret' => 'asdfasgasdfasdasdfa5',
  'default_graph_version' => 'v2.5',
  ]);

$helper = $fb->getRedirectLoginHelper();

$permissions = ['email']; // Optional permissions
$loginUrl = $helper->getLoginUrl('http://toriset.amutepoint.com/fbcallback.php', $permissions);

echo '<a href="' . htmlspecialchars($loginUrl) . '">Log in with Facebook!</a>';
?>

```
And then to make it even more incredibly simple, I don't do ANYTHING in the callback script except echo out the Session variables:

```
<?php
require_once __DIR__ . '/vendor/autoload.php';

session_start();

echo '<pre>';
var_dump($_SESSION);
echo '</pre>';
?>
```

[...]

PR #613, opened 2016-06-14
FacebookRedirectLoginHelper - let's not regenerate CSRF token

KuceraMartin:
When `FacebookRedirectLoginHelper::getLoginUrl()` is called it should first check if a CSRF token (called `state` in the source code) has already been generated and use the existing one if it has.
Right now any time the method is called, a new CSRF token is generated. That means if the method is called more than once, only the last URL will be working which is a problem.

What if the user opens up a webpage in multiple tabs at once? Then he will only be able to use Facebook login in the last tab. And if you forget to put favicon.ico on your website, then he will only be able to use Facebook login in IE :) (see https://github.com/facebook/facebook-php-sdk-v4/issues/583 for details).

### DIFF
--- a/src/Facebook/Helpers/FacebookRedirectLoginHelper.php
+++ b/src/Facebook/Helpers/FacebookRedirectLoginHelper.php
@@ -122,7 +122,7 @@ class FacebookRedirectLoginHelper
      */
     private function makeUrl($redirectUrl, array $scope, array $params = [], $separator = '&')
     {
-        $state = $this->pseudoRandomStringGenerator->getPseudoRandomString(static::CSRF_LENGTH);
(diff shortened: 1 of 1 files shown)


facebook-github-bot-6:
Thank you for your pull request and welcome to our community. We require contributors to sign our Contributor License Agreement, and we don't seem to have you on file. In order for us to review and merge your code, please sign up at https://code.facebook.com/cla - and if you have received this in error or have any questions, please drop us a line at cla@fb.com. Thanks!

facebook-github-bot-8:
Thank you for signing our Contributor License Agreement. We can now accept your code for this (and any) Facebook open source project. Thanks!

SammyK:
This is great stuff! Thanks! This should knock out most of the issues we've been having for a very long time with the CSRF (#583). I'm going to finish this one off by resetting the state after someone successfully obtains an access token so that the same CSRF token isn't being used every time.

PR #628, opened 2016-08-06
Reset the CSRF so that it does not get reused

SammyK:
This finishes off #613


SammyK:
@yguedidi I'm going to go ahead and merge this in since it's just finishing off a partial PR. I'm going to do this with one or two others as well just so we can clean out some PR's and tag. Let me know if you have any issues with that! :)

yguedidi:
Go for that!

Issue #619, opened 2016-07-18
FacebookBatchResponse failed


AidasK:
I was sending loads of successful batch requests when this happened: 
```
...
 if (empty($requests)) {
                continue;
}
$a = $facebook->sendBatchRequest($requests)->getResponses();
```
```
 [Symfony\Component\Debug\Exception\FatalThrowableError]
  Type error: Argument 1 passed to Facebook\FacebookResponse::__construct() must be an instance of Facebook\FacebookRequest, null given, called in /var/www/versions/1468485597/vendor/facebook/php-sdk-v4/src/Facebook/FacebookBatchResponse.php on line 108
```
facebook/php-sdk-v4 5.2.0
php7.0

SammyK:
Hey @AidasK! Can you include a bit more code that shows how you are creating the requests before you send them to `sendBatchRequest()`? :)

yguedidi:
I think the case of timeouted batched request is not handled properly

SammyK:
@yguedidi Ah! Good call. We'll need to look into this. I should be able to give some more TLC in August. :)

AidasK:
I haven't included it, because it is quite straight forward. This error has happened to me only once. I have sent thousands of notifications before and only one time this error occurred. 
```php
$requests = [];
foreach ($ids as $id) {
                $requests[] = $facebook->request('post', "/$id/notifications", [
                    'template' => 'You have a new message',
                    'href' => '?p=' . urlencode($id),
                    'ref' => 'abc',
                ]);
}
 if (empty($requests)) {
                continue;
}
$a = $facebook->sendBatchRequest($requests)->getResponses();
```

SammyK:
Thanks @AidasK! I think @yguedidi nailed it on the head - we need to do better handling of failed batch requests. :)

yguedidi:
@SammyK I'll try ASAP

yguedidi:
@AidasK can you please confirm that the bug happen when you have more that 50 items in your `$ids` array? Thanks

AidasK:
I think I was sending exactly 50 notifications in one batch at a time. But not more, because of FB limitations.

AidasK:
p.s. I don't use facebook notifications anymore, so I don't know if it was fixed. You can close it for now

yguedidi:
@SammyK what do you think if we close this? We can't reproduce it, it's not the more than 50 requests (we already have a safeguard), and it only happened once to @AidasK.

[...]

PR #636, opened 2016-08-19
Add fix for #619

SammyK:
I've tried my darnedest to replicate the issue in #619, but I can't replicate it for the life of me. I was thinking it was because of timeouts where the response will be `null`, but it handles `null` responses just fine.

I'm thinking this has to be an extreme edge case. At any rate, I added this little patch to keep error messages from being generated if the edge case happens again.

@yguedidi - what do you think? :)

### DIFF
--- a/src/Facebook/FacebookBatchResponse.php
+++ b/src/Facebook/FacebookBatchResponse.php
@@ -98,7 +98,7 @@ class FacebookBatchResponse extends FacebookResponse implements IteratorAggregat
     public function addResponse($key, $response)
     {
         $originalRequestName = isset($this->batchRequest[$key]['name']) ? $this->batchRequest[$key]['name'] : $key;
-        $originalRequest = isset($this->batchRequest[$key]['request']) ? $this->batchRequest[$key]['request'] : null;
+        $originalRequest = isset($this->batchRequest[$key]['request']) ? $this->batchRequest[$key]['request'] : new FacebookRequest;
 
         $httpResponseBody = isset($response['body']) ? $response['body'] : null;
         $httpResponseCode = isset($response['code']) ? $response['code'] : null;


yguedidi:
This looks weird to me... I'm :-1: on this solution

The error means that we receive a response that does not have a corresponding request (or at least a corresponding response we do not found).

I think it is better to decide what to do in this case:

- throwing an exception
- allow responses to be created without request

Any preference? :)

SammyK:
I'm cool with throwing an exception. But in what scenario would that happen? I tried to trigger the error by manually throwing invalid values at it but I couldn't trigger it! :)

yguedidi:
OMG!!! @SammyK I think I got it! Not tried yet (late night idea), but I think this can happen due to the limit of 50 requests per batch request. :smile:

SammyK:
That's so awesome! Nice job @yguedidi! And thanks for knocking out some of these issues! I've been pretty swamped lately so I'm looking forward to getting some time to work on the SDK when things let up! :)

src/Facebook/FacebookBatchResponse.php
```diff
@@ -98,7 +98,7 @@ public function setResponses(array $responses)
     public function addResponse($key, $response)
     {
         $originalRequestName = isset($this->batchRequest[$key]['name']) ? $this->batchRequest[$key]['name'] : $key;
-        $originalRequest = isset($this->batchRequest[$key]['request']) ? $this->batchRequest[$key]['request'] : null;
+        $originalRequest = isset($this->batchRequest[$key]['request']) ? $this->batchRequest[$key]['request'] : new FacebookRequest;
```
yguedidi: Instead of a `FacebookRequest`, I'd use an `UnknownFacebookRequest` which extends `FacebookRequest`.
That way, we don't break BC, and the user can check with a simple `instanceof` if the original request is there or not.

Issue #608, opened 2016-06-04
Force the user to choose a Graph API version


yguedidi:
Point 7 in #587

yguedidi:
Closed by #643

PR #643, opened 2016-08-29
Remove default Graph API version fallback

SammyK:
This is a first step to knock out #608. I think in order to close that issue, we'll need to rethink how we're passing around the required vars. You can see I added a hard-coded fallback (temporarily) in `FacebookRequest` since the `$graphVersion` can be `null` and it's at the end of a long list of optional params.

Perhaps we could add a `GraphEnvironment` entity that would include the Graph version, `FacebookApp`, access token, etc. Then we could just pass around the `GraphEnvironment` entity instead of all these individual params. What do you think? :)


src/Facebook/FacebookRequest.php
```diff
@@ -102,7 +102,8 @@ public function __construct(FacebookApp $app = null, $accessToken = null, $metho
         $this->setEndpoint($endpoint);
         $this->setParams($params);
         $this->setETag($eTag);
-        $this->graphVersion = $graphVersion ?: Facebook::DEFAULT_GRAPH_VERSION;
+        // @TODO Remove this fall-back
+        $this->graphVersion = $graphVersion ?: 'v2.7';
```
yguedidi: I think the way to handle this case is to allow null graph version in requests. This parameter is here to override the version given in the client, so it can be null, we just have to take care of this value when sending the request, falling back to the client version when this is null :)

tests/FacebookTest.php
```diff
@@ -36,6 +36,8 @@
 
 class FacebookTest extends \PHPUnit_Framework_TestCase
 {
+    const DEFAULT_GRAPH_VERSION = 'v2.7';
```
yguedidi: should be changed to 2.8 :)

src/Facebook/FacebookRequest.php
mozarliker: Hello yguedidi! I have no certainties about this case version 2.8 yet and icoguinita for me, but I resolve to be with this team for sure has brilhates minds, we try faser tests, for and studying the case to come to perfection. I trust you guys the importate and not waste time, bad weather do quick pass worth every second. with an impeccable work. ok let's see what happens reversing from 2.7 to 2.8 ok

SammyK: @mozarliker

SammyK:
Sorry for the delay @yguedidi! I've made the changes you requested and also updated the tests. :) Do you have any other thoughts on this one?

tests/FacebookRequestTest.php
```diff
@@ -125,14 +125,14 @@ public function testAProperUrlWillBeGenerated()
...
 
         $this->assertEquals($expectedUrl, $getUrl);
 
         $postRequest = new FacebookRequest($app, 'foo_token', 'POST', '/bar', ['foo' => 'bar']);
 
         $postUrl = $postRequest->getUrl();
-        $expectedUrl = '/' . Facebook::DEFAULT_GRAPH_VERSION . '/bar';
+        $expectedUrl = '/bar';
 
         $this->assertEquals($expectedUrl, $postUrl);
```
yguedidi: I'd add a case for a versionned request, as it useful in the case of batched request to use a different version than the one of the master request :)

tests/FacebookTest.php
```diff
@@ -36,9 +36,12 @@
 
 class FacebookTest extends \PHPUnit_Framework_TestCase
 {
+    const DEFAULT_GRAPH_VERSION = 'v2.8';
```
yguedidi: I'd remove this constant and hardcode it's value everywhere it's used. In tests, we do not need to maintain a consistent value across all tests.
BTW, that why I'd change the value to `vX.X` or `v4.2` (haha), remember to change the value each time the Graph API gets updated is a pain.

SammyK:
Thanks as always @yguedidi! :)

PR #650, opened 2016-09-07
Fix: Method does not return anything

localheinz:
This PR

* [x] updates a docblock for a method that doesn't return anything


SammyK:
Thanks! :)

PR #662, opened 2016-10-05
Made AccessToken::isLongLived conform to the documentation.

smatyas:
The `AccessToken::isLongLived` method is not conform to the documentation.
This PR fixes this issue by modifying the returned value in this particular case, to meet the docs.

The doc says:

>If the expiration date was not originally provided, the method will return `null`.

https://github.com/facebook/php-graph-sdk/blob/master/docs/AccessToken.fbmd#L31

### DIFF
--- a/src/Facebook/Authentication/AccessToken.php
+++ b/src/Facebook/Authentication/AccessToken.php
@@ -93,7 +93,7 @@ class AccessToken
     /**
      * Determines whether or not this is a long-lived token.
      *
-     * @return bool
+     * @return bool|null
      */
     public function isLongLived()
     {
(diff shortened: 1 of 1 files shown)


SammyK:
Thanks for the contribution @smatyas! Since this would be a breaking change, we should update the docs instead of the codebase. If you'd like to send another PR that updates the docs, that'd be really appreciated! :)

smatyas:
Yeah, I was thinking about the BC break too, and I agree. 
Although I would me more happy if this method would work the same way as the `isExpired` for example. Maybe in the next major :)

PR #676, opened 2016-10-14
Docs to md (first round)

SammyK:
Refactored the landing page of the docs to standard markdown. We'll probably want to organize these into folders: `examples/` and `reference/`.


docs/README.md
```diff
@@ -0,0 +1,34 @@
+# Facebook SDK v5 for PHP
```
yguedidi: I'd just remove the v5 :)

yguedidi:
Totally agree for folders!

PR #678, opened 2016-10-18
fixed invalid format handling of headers[x-fb-ads-insights-throttle]

tituskeuler:
The raw response header field "X-FB-Ads-Insights-Throttle" contains a json value { "app_id_util_pct": 100, "acc_id_util_pct": 10 }

Calling $response->getHeaders()['x-fb-ads-insights-throttle'] results in a chopped value {"app_id_util_pct".

The fix will handle the transformation of the header fields in a valid way. 

Similar to this fix:
https://github.com/facebook/facebook-php-ads-sdk/pull/241/files

### DIFF
--- a/src/Facebook/Http/GraphRawResponse.php
+++ b/src/Facebook/Http/GraphRawResponse.php
@@ -129,7 +129,7 @@ class GraphRawResponse
             if (strpos($line, ': ') === false) {
                 $this->setHttpResponseCodeFromHeader($line);
             } else {
-                list($key, $value) = explode(': ', $line);
+                list($key, $value) = explode(': ', $line, 2);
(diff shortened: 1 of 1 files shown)


SammyK:
Nice catch & thanks for the PR @tituskeuler! Would you mind adding a test for this as well? :)

yguedidi:
@tituskeuler this PR should be closed and reopened targetting the 5.4 branch as master is now 6.0 :)

yguedidi:
And `GraphRawResponse` class will be removed in 6.0 due to HTTPlug

tituskeuler:
Ok, I did not see that. So I close this one and create the PR against 5.4. I will add the test in that request then. CLA was signed.

PR #679, opened 2016-10-18
Fixed invalid format handling of headers[x-fb-ads-insights-throttle]

tituskeuler:
The raw response header field "X-FB-Ads-Insights-Throttle" contains a json value { "app_id_util_pct": 100, "acc_id_util_pct": 10 }

Calling $response->getHeaders()['x-fb-ads-insights-throttle'] results in a chopped value {"app_id_util_pct".

The fix will handle the transformation of the header fields in a valid way.

Similar to this fix:
https://github.com/facebook/facebook-php-ads-sdk/pull/241/files

Relates to https://github.com/facebook/php-graph-sdk/pull/678


SammyK:
Thanks so much for the contribution @tituskeuler! :)

PR #680, opened 2016-10-18
Enhancement: Combined guzzle handler checks

bibumathew:


SammyK:
Thanks for the PR @bibumathew! I'd like to keep it the way it currently because embedded `if` statements are harder to read. I'll let @yguedidi have the final say though. :)

yguedidi:
Thanks @bibumathew! I agree with @SammyK to keep it like it is, as it's "handle errors first, then returns the right value".
BTW, you target the wrong branch, as master is now 6.0 and HTTP clients will be removed in this version :)

bibumathew:
Thank you

PR #698, opened 2016-11-08
Update SignedRequest.php

Incont:
Changing for representation "payment_id" as string, because "payment_id" has 15 decimal digits and float type in php has 14 decimal digits so 1 digit is lost


yguedidi:
According [this reference page](https://developers.facebook.com/docs/reference/login/signed-request), there is no `payment_id` field in signed request. Any reference?

yguedidi:
Closed as duplicate of #699

PR #699, opened 2016-11-08
Update SignedRequest.php

Incont:
Changing for representation "payment_id" as string, because "payment_id" has 15 decimal digits and float type in php has 14 decimal digits so 1 digit is lost. (increase depth to max like default)

### DIFF
--- a/src/Facebook/SignedRequest.php
+++ b/src/Facebook/SignedRequest.php
@@ -211,7 +211,7 @@ class SignedRequest
         $payload = $this->base64UrlDecode($encodedPayload);
 
         if ($payload) {
-            $payload = json_decode($payload, true);
+            $payload = json_decode($payload, true, 512, JSON_BIGINT_AS_STRING);
         }
 
         if (!is_array($payload)) {


yguedidi:
According [this reference page](https://developers.facebook.com/docs/reference/login/signed-request), there is no `payment_id` field in signed request. Any reference?

Incont:
I tried to use this class for working with "signed_request" from payer. https://developers.facebook.com/docs/games_payments/fulfillment (3. order placement)

SammyK:
Great stuff @Incont! :) Since Facebook ID's are massive in general, I'd love to see this update get added to [FacebookResponse::decodeBody()](https://github.com/facebook/php-graph-sdk/blob/master/src/Facebook/FacebookResponse.php#L236) as well. But probably for v6 since that's technically a BC.

@yguedidi You see any issues with merging this into master and cherry-picking to 5.4?

yguedidi:
Most of modern servers are 64bits, which supports bigint (see https://3v4l.org/57gIj), so this can be considered as a bugfix for old 32bit servers, so I'm OK :)

SammyK:
Thanks @Incont! :)

PR #706, opened 2016-12-03
Implement omit_response_on_success batch option

teldosas:
This adresses #438 implementing the interface proposal of @SammyK . I haven't updated the docs yet. I will do so after the review.

### DIFF
--- a/docs/examples/batch_request.md
+++ b/docs/examples/batch_request.md
@@ -96,7 +96,65 @@ My next 2 events are House Warming Party,Some Foo Event.
 
 It should also contain a response containing two photos from the user.
 
-> **Warning:** The response object should return a `null` response for any request that was pointed to with JSONPath as is [the behaviour of the batch functionality of the Graph API](https://developers.facebook.com/docs/graph-api/making-multiple-requests/#operations).
+> **Warning:** The response object should return a `null` response for any request that was pointed to with JSONPath as is [the behaviour of the batch functionality of the Graph API](https://developers.facebook.com/docs/graph-api/making-multiple-requests/#operations). If we want to receive the response anyway we have to set the `omit_response_on_success` option to `false`. [See the example below](#force-response-example).
+
+## Force Response Example
+
+The following example is a subset of the [first example](#example). We will only use the `user-events` and `post-to-feed` requests of the [first example](#example), but in this case we will force the server to return the response of the `user-events` request.
+
+```php
+$fb = new Facebook\Facebook([
+  'app_id' => '{app-id}',
+  'app_secret' => '{app-secret}',
+  'default_graph_version' => 'v2.8',
+  ]);
+
+// Since all the requests will be sent on behalf of the same user,
+// we'll set the default fallback access token here.
+$fb->setDefaultAccessToken('user-access-token');
+
+// Get user events
+$requestUserEvents = $fb->request('GET', '/me/events?fields=id,name&limit=2');
+
+// Post a status update with reference to the user's events
+$message = 'My next 2 events are {result=user-events:$.data.*.name}.';
+$statusUpdate = ['message' => $message];
+$requestPostToFeed = $fb->request('POST', '/me/feed', $statusUpdate);
+
+// Create an empty batch request
+$batch = $fb->newBatchRequest();
+
+// Populate the batch request
+// Set the 'omit_response_on_success' option to false to force the server return the response
+$batch->add($requestUserEvents, "user-events", ["omit_response_on_success" => false]);
+$batch->add($requestPostToFeed, "post-to-feed");
+
+// Send the batch request
+try {
+  $responses = $fb->getClient()->sendBatchRequest($batch);
+} catch(Facebook\Exceptions\FacebookResponseException $e) {
+  // When Graph returns an error
+  echo 'Graph returned an error: ' . $e->getMessage();
+  exit;
+} catch(Facebook\Exceptions\FacebookSDKException $e) {
+  // When validation fails or other local issues
+  echo 'Facebook SDK returned an error: ' . $e->getMessage();
+  exit;
+}
+
+foreach ($responses as $key => $response) {
+  if ($response->isError()) {
+    $e = $response->getThrownException();
+    echo '<p>Error! Facebook SDK Said: ' . $e->getMessage() . "\n\n";
+    echo '<p>Graph Said: ' . "\n\n";
+    var_dump($e->getResponse());
+  } else {
+    echo "<p>(" . $key . ") HTTP status code: " . $response->getHttpStatusCode() . "<br />\n";
+    echo "Response: " . $response->getBody() . "</p>\n\n";
+    echo "<hr />\n\n";
+  }
+}
+```
 
 ## Multiple User Example
 
--- a/docs/examples/batch_request.md
+++ b/docs/examples/batch_request.md
@@ -156,6 +156,71 @@ foreach ($responses as $key => $response) {
 }
 ```
 
+## Explicit Dependency Example
+
+In the following example we will make two requests.
+* One to post a status update on the user's feed
+* and one to receive the last post of the user (which should be the one that we posted with first request).
+
+Since we want the second request to be executed after the first one is completed, we have to set the `depends_on` option of the second request to point to the name of the first request. We assume that we have the following options granted from the user: `user_posts`, `publish_actions`.
+
+```php
+$fb = new Facebook\Facebook([
+  'app_id' => '{app-id}',
+  'app_secret' => '{app-secret}',
+  'default_graph_version' => 'v2.8',
+  ]);
+
+// Since all the requests will be sent on behalf of the same user,
+// we'll set the default fallback access token here.
+$fb->setDefaultAccessToken('user-access-token');
+
+// Post a status update to the user's feed
+$message = 'Random status update';
+$statusUpdate = ['message' => $message];
+$requestPostToFeed = $fb->request('POST', '/me/feed', $statusUpdate);
+
+// Get last post of the user
+$requestLastPost = $fb->request('GET', '/me/feed?limit=1');
+
+// Create an empty batch request
+$batch = $fb->newBatchRequest();
+
+// Populate the batch request
+$batch->add($requestPostToFeed, "post-to-feed");
+
+// Set the 'depends_on' property to point to the first request
+$batch->add($requestLastPost, "last-post", ["depends_on" => "post-to-feed"]);
+
+// Send the batch request
+try {
+  $responses = $fb->getClient()->sendBatchRequest($batch);
+} catch(Facebook\Exceptions\FacebookResponseException $e) {
+  // When Graph returns an error
+  echo 'Graph returned an error: ' . $e->getMessage();
+  exit;
+} catch(Facebook\Exceptions\FacebookSDKException $e) {
+  // When validation fails or other local issues
+  echo 'Facebook SDK returned an error: ' . $e->getMessage();
+  exit;
+}
+
+foreach ($responses as $key => $response) {
+  if ($response->isError()) {
+    $e = $response->getThrownException();
+    echo '<p>Error! Facebook SDK Said: ' . $e->getMessage() . "\n\n";
+    echo '<p>Graph Said: ' . "\n\n";
+    var_dump($e->getResponse());
+  } else {
+    echo "<p>(" . $key . ") HTTP status code: " . $response->getHttpStatusCode() . "<br />\n";
+    echo "Response: " . $response->getBody() . "</p>\n\n";
+    echo "<hr />\n\n";
+  }
+}
+```
+
+> **Warning:** The response object should return a `null` response for any request that was pointed to with the `depends_on` option as is [the behaviour of the batch functionality of the Graph API](https://developers.facebook.com/docs/graph-api/making-multiple-requests/#operations). If we want to receive the response anyway we have to set the `omit_response_on_success` option to `false`. [See example](#force-response-example).
+
 ## Multiple User Example
 
 Since the requests sent in a batch are unrelated by default, we can make requests on behalf of multiple users and pages in the same batch request.
--- a/src/Facebook/Facebook.php
+++ b/src/Facebook/Facebook.php
@@ -482,12 +482,14 @@ class Facebook
         return $this->lastResponse = $this->client->sendBatchRequest($batchRequest);
     }
 
+
     /**
      * Instantiates an empty FacebookBatchRequest entity.
      *
      * @param  AccessToken|string|null $accessToken  The top-level access token. Requests with no access token
      *                                               will fallback to this.
-     * @param  string|null $graphVersion The Graph API version to use.
+     * @param  string|null             $graphVersion The Graph API version to use.
+     *
      * @return FacebookBatchRequest
      */
     public function newBatchRequest($accessToken = null, $graphVersion = null)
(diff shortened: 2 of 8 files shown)


src/Facebook/FacebookBatchRequest.php
```diff
@@ -250,7 +257,10 @@ public function requestEntityToBatchArray(FacebookRequest $request, $requestName
             $batch['attached_files'] = $attachedFiles;
         }
 
-        // @TODO Add support for "omit_response_on_success"
+        if (isset($options['omit_response_on_success'])) {
```
SammyK: I'm thinking we just merge the options straight in so we can support all the existing options plus future ones.

```php
if (!empty($options)) {
    $batch = array_merge($options, $batch);
}
```

...or `array_merge($batch, $options)`, but that would allow the users to break things if they passed in an option named `headers` with random stuff in it for example.

But this makes me think we could get rid of `$requestName` and just pass in `name` through `$options`. Hum..

@yguedidi thoughts? :)

yguedidi: I think `name`, `attached_files` and current `options` can be merge in one `options` parameter.
And to resolve your issue with merging @SammyK: `$batch += $options;` ;)

teldosas:
@SammyK  @yguedidi I've made the changes :)

src/Facebook/Facebook.php
```diff
@@ -483,6 +483,27 @@ public function sendBatchRequest(array $requests, $accessToken = null, $graphVer
     }
 
     /**
+     * Instantiates an empty FacebookBatchRequest entity.
+     *
+     * @param  AccessToken|string|null $accessToken  The top-level access token. Requests with no access token
+     *                                               will fallback to this.
+     * @param  string|null $graphVersion The Graph API version to use.
```
yguedidi: wrong description alignment :)
BTW, I'd add an empty line between this `@param` and the below `@return`

src/Facebook/FacebookBatchRequest.php
```diff
@@ -62,16 +62,17 @@ public function __construct(FacebookApp $app = null, array $requests = [], $acce
     }
 
     /**
-     * A a new request to the array.
+     * Adds a new request to the array.
      *
      * @param FacebookRequest|array $request
      * @param string|null           $name
+     * @param array                 $options The batch options e.g. 'omit_response_on_success'
```
yguedidi: I'd typehint it as `string[]`

src/Facebook/FacebookBatchRequest.php
```diff
@@ -91,6 +92,10 @@ public function add($request, $name = null)
             'request' => $request,
         ];
 
+        if (!empty($options)) {
+            $requestToAdd['options'] = $options;
```
yguedidi: I'd move this in the array initialization above, no need to check for empty, it will always be an array thanks to the typehint

src/Facebook/FacebookBatchRequest.php
```diff
@@ -189,8 +194,19 @@ public function convertRequestsToJson()
     {
         $requests = [];
         foreach ($this->requests as $request) {
-            $attachedFiles = isset($request['attached_files']) ? $request['attached_files'] : null;
-            $requests[] = $this->requestEntityToBatchArray($request['request'], $request['name'], $attachedFiles);
+            $options = [];
+
+            if (isset($request['name'])) {
```
yguedidi: `name` is always set, I'd check for `!empty($request['name'])`

src/Facebook/FacebookBatchRequest.php
```diff
@@ -216,12 +232,12 @@ public function validateBatchRequestCount()
      * Converts a Request entity into an array that is batch-friendly.
      *
      * @param FacebookRequest $request       The request entity to convert.
```
yguedidi: wrong description alignment

src/Facebook/FacebookBatchRequest.php
```diff
@@ -189,8 +194,19 @@ public function convertRequestsToJson()
...
+
+            if (isset($request['name'])) {
+                $options['name'] = $request['name'];
+            }
+
+            if (isset($request['attached_files'])) {
+                $options['attached_files'] = $request['attached_files'];
+            }
+
+            $options += isset($request['options']) ? $request['options'] : [];
```
yguedidi: when my above comment about always set the `options` key, this become just `$options += $request['options'];` :)

src/Facebook/FacebookBatchRequest.php
```diff
@@ -189,8 +194,19 @@ public function convertRequestsToJson()
...
         foreach ($this->requests as $request) {
-            $attachedFiles = isset($request['attached_files']) ? $request['attached_files'] : null;
-            $requests[] = $this->requestEntityToBatchArray($request['request'], $request['name'], $attachedFiles);
+            $options = [];
+
+            if (isset($request['name'])) {
+                $options['name'] = $request['name'];
+            }
+
+            if (isset($request['attached_files'])) {
```
yguedidi: like the `options` key, I'll always set the `attached_files` key in `add()`, and then check here for `!empty($request['attached_files'])`

teldosas:
@yguedidi I made these changes too! :) The last review comment was not outdated even though I did force push and fixed the alignment. I'll start with the docs asap. ;)

teldosas:
BTW `options`  and `attached_files` being set always cause 4 tests to fail because they use the method `getRequests()` which returns the `requests` array right after being processed from the `add()` method and the array contains two extra keys with empty or null values.

Also checking `!empty($request['name'])` before setting the name does not add the name of the firsts request when the requests are added as a numeric array because `0` is considered to be empty.

src/Facebook/FacebookBatchRequest.php
```diff
@@ -189,8 +191,19 @@ public function convertRequestsToJson()
     {
         $requests = [];
         foreach ($this->requests as $request) {
-            $attachedFiles = isset($request['attached_files']) ? $request['attached_files'] : null;
-            $requests[] = $this->requestEntityToBatchArray($request['request'], $request['name'], $attachedFiles);
+            $options = [];
+
+            if (!empty($request['name'])) {
```
yguedidi: To solve the issues, check here for `null !== $request['name']`, and in the `add` function, cast the `$key` to a string :)

src/Facebook/FacebookBatchRequest.php
```diff
@@ -189,8 +191,19 @@ public function convertRequestsToJson()
...
+                $options['name'] = $request['name'];
+            }
+
+            if (!empty($request['attached_files'])) {
```
yguedidi: BTW, you can change this too to `null !== $request['attached_files']` as `extractFileAttachments` always returns a string or null

yguedidi:
Thanks @teldosas! I added some comments to fix your issues. About tests, just change the expected value in thoses failing tests :)

src/Facebook/FacebookBatchRequest.php
teldosas: @yguedidi Then why not just use the `isset()` function that returns false when the variable is null. Plus I think that this is how we check in similar cases throughout the project.

src/Facebook/FacebookBatchRequest.php
yguedidi: Updated, I'd added using `strval()`

src/Facebook/FacebookBatchRequest.php
yguedidi: @teldosas "`extractFileAttachments` always returns a **non empty** string or null", sorry I missed an important part :)

`isset()` will of course work, but sementically it does not make sense, because the `attached_files` key is always set.
BTW, I tend to be strict about types: `null` means no attached files/no name, a string value means I want a value. 

Then the question is what to do when that value is invalid (e.g. an empty string).

I think we have two options :
- throwing an exception
- let Facebook Graph API returns an error

I personally prefer the last option. It's tedious to handle **all** user's type/value errors, we already pretty well document the usage of our functions :)

ping @SammyK, thoughts?

src/Facebook/FacebookBatchRequest.php
yguedidi: @teldosas I'm talking about the `$key` variable in `add()`, the case when you give an array of requests.
In `->add($myRequest)`, `$name` will be null
In `->add($ManyRequests)`, `$name` for eachrequest will be its key in the array, an integer, which is not allowed by our doc, that's why I suggest you to `->add($request, strval($key))` in the loop :)

teldosas: Ok it makes sense now, thanks :)

yguedidi:
Just one more thing, this PR targets master, which is the v6 dev branch, and this makes sense as it introduce a small BC break.
But it introduce a great addition that can be added without BC break, so what about closing this PR and open a new one that targets 5.4 and remove the BC break? :)

ping @SammyK @teldosas (sorry for making you wait again)

teldosas:
I added the docs and I have some notes/questions:
1. After playing a bit around with this new feature I realized that this PR does not add support for `JSONP callback`. Furthermore I don't think that the **`JSONP callback`** can or should be implemented inside the `requestEntityToBatchArray` or the `add` method because it is **general option of every request**.
For example in a simple request one would add a `JSONP callback` like this: 
     ```php
     $req = $fb->get('/me/feed?callback'={function_name});
     ```
     (just by adding a `callback` get/post parameter).
But in the case of a `FacebookBatchRequest` we don't have anyway of adding a form post parameter. Thus I propose that we open a new issue to discuss an interface for adding a `JSONP callback` to a batch request. Maybe add a method 
     ```php 
     FacebookBatchRequest::setJSONPCallback(string $functionName)
     ```
     or even make it more general
     ```php 
     FacebookRequest::setJSONPCallback(string $functionName)
     ```
    Thoughts?

2. @yguedidi I am totally fine with the new PR but you would have to enlighten me about what this small Backwards Compatibility break is (I have no idea). :)

3. Is it a problem that we provide two (or maybe even three) ways for a user to create and send a batch request and only one of those supports the batch options (the one implemented in this PR)?

docs/examples/batch_request.md
```diff
@@ -96,7 +96,130 @@ My next 2 events are House Warming Party,Some Foo Event.
...
+
+// Populate the batch request
+$batch->add($requestPostToFeed, "post-to-feed");
+
+// Set the 'depends_on' property to point to the first request
+$batch->add($requestLastPost, "last-post", ["depends_on" => "post-to-feed"]);
+
+// Send the batch request
+try {
+  $responses = $fb->getClient()->sendBatchRequest($batch);
```
teldosas: Since this part that sends and prints the batch request is repeated in the four examples of this file, could we maybe add a new section here about sending a batch request? Maybe with two subsections:
* one for sending an array of `Facebook\FacebookRequest`s
* and one for sending an instance of `Facebook\FacebookBatchRequest`

yguedidi: I agree, but this should be in a separate PR ;)

src/Facebook/FacebookBatchRequest.php
```diff
@@ -216,12 +229,12 @@ public function validateBatchRequestCount()
...
      * @param FacebookRequest $request       The request entity to convert.
-     * @param string|null     $requestName   The name of the request.
-     * @param string|null     $attachedFiles Names of files associated with the request.
+     * @param array           $options       The name and the attached files of the request plus
+     *                                       the batch options e.g. 'omit_response_on_success'
      *
      * @return array
      */
-    public function requestEntityToBatchArray(FacebookRequest $request, $requestName = null, $attachedFiles = null)
+    public function requestEntityToBatchArray(FacebookRequest $request, array $options = [])
```
yguedidi: Here is the BC break, because `requestEntityToBatchArray` is public, changing its signature is a BC break. :)

yguedidi:
@teldosas Thanks for the doc, excellent!

1. Why do you want to use JSONP in a PHP SDK? Your PHP code will not execute the JSONP response :)
2. See my comment in the code for the BC break. BTW, if you want to learn more about what breaks, Symfony framework has a good doc about it: http://symfony.com/doc/current/contributing/code/bc.html#working-on-symfony-code
3. Not for me, as we will work on v6. @SammyK thoughts?

SammyK:
@teldosas Thanks for all the hard work on this! :)

1 & 2: +1 to what @yguedidi said :)
3: Let's implement this way first that supports all the options and if we want to create a facade on the `Facebook` class make batch requests easier, we can, but let's do that in another issue/PR. :)

src/Facebook/FacebookBatchRequest.php
```diff
@@ -62,20 +62,21 @@ public function __construct(FacebookApp $app = null, array $requests = [], $acce
...
      * @param FacebookRequest|array $request
      * @param string|null           $name
+     * @param string[]              $options The batch options e.g. 'omit_response_on_success'
      *
      * @return FacebookBatchRequest
      *
      * @throws \InvalidArgumentException
      */
-    public function add($request, $name = null)
+    public function add($request, $name = null, array $options = [])
```
yguedidi: BTW, this is a BC break too in case someone extends `FacebookBatchRequest`

SammyK: Yeah, I'm curious if we can even add this to 5.4 without adding a BC break. Can you think of a way outside of adding a new method for "adding with options" or something?

yguedidi: It looks a bit tricky, but it works! Renaming `$name` to `$options` and document that if it's a string it will be the request name, otherwise it will be the array of options :)

yguedidi: @SammyK BC is something hard to keep, thats why we have to have the closest public API for v6, it ease adding features without break things :)

SammyK: Love it! Oh the wonky things we can get away with thanks to duck typing... :)

teldosas: So do we merge this to master and start a new PR for 5.4 with the aforementioned interface, or do we discard this PR and do just the PR for 5.4?

Also can I rename this to `Add full batch support` ?

src/Facebook/FacebookBatchRequest.php
```diff
@@ -62,20 +62,21 @@ public function __construct(FacebookApp $app = null, array $requests = [], $acce
...
     {
         if (is_array($request)) {
             foreach ($request as $key => $req) {
-                $this->add($req, $key);
+                $this->add($req, strval($key));
```
SammyK: @teldosas Why did we add `strval()` here?
@yguedidi This is a BC break when you're expecting numeric array keys since they will be returned as strings now.

SammyK:
> So do we merge this to master and start a new PR for 5.4 with the aforementioned interface, or do we discard this PR and do just the PR for 5.4?

I'd say let's close this PR and do a new PR that targets the 5.4 branch. Then I can merge it into master. Once merged into master, we can create a new PR with any BC breaks we want to do. :)

> Also can I rename this to Add full batch support ?

Sure!

> (About the JSONP) It makes sense. This @TODO tag was confusing ;)

Yeah, I remember adding that a while ago and then forgetting to remove it. :)

src/Facebook/FacebookBatchRequest.php
yguedidi: @SammyK I tell @teldosas to add this. `add()` expects a string for the name, let's give it a string :)
How it can break?

teldosas:
I started the PR for 5.4 :) #713

SammyK:
Close since this was addressed in #713. Thanks! :)

PR #710, opened 2016-12-15
Fix the exception message on docs example

alexndreazevedo:


SammyK:
Thanks @alexndreazevedo! :)

alexndreazevedo:
You're welcome, @SammyK

PR #712, opened 2016-12-23
Add "Security Vulnerabilities" section to README

SammyK:
In case there are new security issues we don't want to have publicly exposed via the issues. I put my email but if @yguedidi wants to be the contact, I'm totally cool with it! :)


SammyK:
Thanks! :)

PR #720, opened 2017-01-06
Enhancement: Keep packages sorted without specifying --sort-packages

localheinz:
This PR

* [x] keeps packages in `composer.json` sorted without the need to specify the `--sort-packages` flag when requiring packages

💁‍♂️ For reference, see https://getcomposer.org/doc/06-config.md#sort-packages.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,14 +11,17 @@
             "homepage": "https://github.com/facebook/php-graph-sdk/contributors"
         }
     ],
+    "config": {
+        "sort-packages": true
+    },
     "require": {
         "php": "^5.6 || ^7.0",
         "paragonie/random_compat": "^2.0"
     },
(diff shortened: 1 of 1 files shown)


SammyK:
Hey @localheinz! Since I merged #728, this one has a conflict.

@yguedidi I have no strong opinions on this one way or another. Your call. :)

localheinz:
@SammyK @yguedidi 

Rebased!

yguedidi:
According the doc, this option is for `require` command, and this command is mostly used in business project, and almost never used in the case of a library... So I am against the config addition.

About ordering dev requirements, why not, so it's your call now @SammyK :)

SammyK:
Why not. Thanks @localheinz! :)

PR #721, opened 2017-01-06
Fix: Run build on PHP7.1

localheinz:
This PR

* [x] adjusts `.travis.yml` to run a build on PHP7.1 (instead of the RC)
* [x] requires builds on PHP7.1 to pass

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,3 @@ install:
 script:
   - vendor/bin/phpcs
   - vendor/bin/phpunit --coverage-text --exclude-group integration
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -25,4 +25,4 @@ script:
 
 matrix:
   allow_failures:
```
SammyK: We can remove `matrix.allow_failures` since 7.1 is stable release now. :)

localheinz:
Thank you, @yguedidi!

PR #723, opened 2017-01-07
Fix: Remove useless override

localheinz:
This PR

* [x] removes a method that needlessly overrides a method in the parent class

💁‍♂️ The code is exactly the same, see [`Facebook\GraphNodes\GraphNodeFactory::makeGraphEvent()`](https://github.com/facebook/php-graph-sdk/blob/master/src/Facebook/GraphNodes/GraphNodeFactory.php#L158-L168).


localheinz:
Thank you, @yguedidi!

PR #724, opened 2017-01-07
Fix: Provide more specific docblock

localheinz:
This PR

* [x] provides more specific type information on a field as well as an accessor


localheinz:
Thank you, @yguedidi!

PR #725, opened 2017-01-07
Fix: Remove unused imports

localheinz:
This PR

* [x] removes unused imports

💁‍♂️ Ran

```
$ php-cs-fixer fix --using-cache=no --rules=no_unused_imports tests
```


localheinz:
Thank you, @yguedidi!

PR #726, opened 2017-01-07
Fix: Use more appropriate assertions

localheinz:
This PR

* [x] uses more appropriate assertions


localheinz:
Thank you, @yguedidi!

PR #727, opened 2017-01-07
Fix: Use class keyword

localheinz:
This PR

* [x] makes use of the `class` keyword instead of specifying class names by string literals


yguedidi:
Thank you so much @localheinz!! How did you do that? Is there a tool that do that automatically? :)
@SammyK what do you think about this one?

localheinz:
@yguedidi 

I uses @kalessil's PhpStorm plugin (see https://github.com/kalessil/phpinspectionsea) and then sorted out a few imports!

SammyK:
I'm a huge fan of this since it makes refactoring a breeze with a good IDE. I'm glad the SDK can finally take advantage of this feature since the min PHP version will be 5.6. :) Thanks @localheinz!

PR #729, opened 2017-01-07
Fix: Apply PSR-2 to credentials template

localheinz:
This PR 

* [x] applies PSR-2 to the test credentials template file


localheinz:
Thank you, @yguedidi!

PR #730, opened 2017-01-07
Fix: Use double pipe for specifying alternate PHP versions

localheinz:
This PR

* [x] uses `||` when specifying platform requirements

💁‍♂️ While `|` is also supported, the documentation suggests to use `||`. For reference, see https://getcomposer.org/doc/articles/versions.md#range.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
         }
     ],
     "require": {
-        "php": "^5.6|^7.0",
+        "php": "^5.6 || ^7.0",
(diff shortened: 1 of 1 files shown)


localheinz:
Thank you, @yguedidi!

SammyK:
Thanks for all the TLC you've been giving to the PHP SDK @localheinz! And thanks for reviewing and merging @yguedidi! :)

PR #740, opened 2017-01-28
Use filter_var instead of regex (preg_match)

J2TeaM:


yguedidi:
@J2TeaM Thanks! But this is a wrong change. `filter_var()` will allow ANY valid URL, and for example, `file:///var/usr/foo` is a valide one, but it isn't a remote file (see [here](https://3v4l.org/ubd6L)) ;)

PR #742, opened 2017-01-28
Cache size of the array

J2TeaM:


yguedidi:
@J2TeaM thank you for your contribution! I'm not a big fan of micro-optimization, especially in this case because `count()` has an O(1) complexity, so caching isn't necessary IMO

BTW, the variable naming is confusing, it's not really a length, it's the date part count, so the variable should be named `$partCount`.
Then, is it really necessary to change the code? :)

PR #766, opened 2017-04-14
Add PHP 7.1 to Travis build

emirb:
I have added 7.1 to Travis build and marked the mcrypt random string test as skipped for >=7.1 because it will never be returned anyway (check https://github.com/facebook/php-graph-sdk/blob/5.4/src/Facebook/PseudoRandomString/PseudoRandomStringGeneratorFactory.php#L80).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ php:
   - 5.5
   - 5.6
   - 7.0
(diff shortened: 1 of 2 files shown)


SammyK:
Thanks @emirb! :)

PR #773, opened 2017-04-28
Generate Better OpCode

tolbon:
I add '\\' to all functions in PHP cause generate better opcode for PHP VM in PHP 7.
Replace a call to INIT_NS_FCALL_BY_NAME to a direct call.

For example : 
If you use strlen in namespace Foo\Bar
PHP search for Foo\Bar\strlen, failed, search in Foo\strlen, failed, search in global namespace and find strlen function.
with '\\' PHP directly ask to a global namespace the function strlen.

### DIFF
--- a/src/Facebook/Authentication/AccessToken.php
+++ b/src/Facebook/Authentication/AccessToken.php
@@ -67,7 +67,7 @@ class AccessToken
      */
     public function getAppSecretProof($appSecret)
     {
-        return hash_hmac('sha256', $this->value, $appSecret);
+        return \hash_hmac('sha256', $this->value, $appSecret);
     }
 
     /**
@@ -87,7 +87,7 @@ class AccessToken
      */
     public function isAppAccessToken()
     {
-        return strpos($this->value, '|') !== false;
+        return \strpos($this->value, '|') !== false;
     }
 
     /**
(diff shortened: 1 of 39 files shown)


yguedidi:
Thanks @tolbon for your contribution, and sorry for the delay...
I disagree with you suggestion, it makes the code less clear to me.
I think the benefit is negligeable, maybe you could profive a benchmark?

tolbon:
Hi @yguedidi, I am agree with you :  

> it makes the code less clear.

I see 2 conferences : 
one who say "do this"
and an other, after my comit, who say "Do this if you want, it's very tiny impact on perfomance"
If you want links (french conferences) ?

So with tiny impact vs readability I suggest you can refuse this PR :+1:

yguedidi:
I'd like the links yes, I'm french :)

And just to quote @Ocramius from [another PR](https://github.com/facebook/php-graph-sdk/pull/641#discussion_r123912344):

> Also: performance in a thing that does HTTP requests is kinda useless anyway ;-)

So I'm closing this, thanks for the contribution

tolbon:
https://youtu.be/AVnkYJn4YwE?t=7m25s
in the same conf at 32:25 we can use encapsed string optim of PHP7, impact less than 10 functions in sdk in my remember.

for the second link sorry impossible to find :/

yguedidi:
Thanks!

PR #777, opened 2017-05-07
Added stop_time to the shouldCastAsDateTime whitelist array

### LINKED ISSUE #774: Field "stop_time" is not cast to DateTime
I am reading Campaign Objects from the Marketing API:

https://developers.facebook.com/docs/marketing-api/reference/ad-campaign-group

These Objects contain (among others) the fields `created_time`, `updated_time`, `start_time` and `stop_time`. All of them are returned as DateTime Objects from the Graph SDK, except the `stop_time`, which is a string.

I would expect the `stop_time` to be a DateTime object as well.

I believe the responsible code is this:
https://github.com/facebook/php-graph-sdk/blob/master/src/Facebook/GraphNodes/GraphNode.php#L136-L155

lpavlovi:
Fix #774

### DIFF
--- a/src/Facebook/GraphNodes/GraphNode.php
+++ b/src/Facebook/GraphNodes/GraphNode.php
@@ -146,6 +146,7 @@ class GraphNode extends Collection
             'created_time',
             'updated_time',
             'start_time',
+            'stop_time',
             'end_time',
             'backdated_time',
             'issued_at',


tolbon:
Just for information, in which node do you have stop_time field ?

tolbon:
My bad I didn't see #774 issues but you didn't use specific sdk for this ?

lpavlovi:
Intended for version 6 of the SDK as of now.

yguedidi:
@lpavlovi Thanks for your contribution!
Just one question, shouldn't you use the facebook/facebook-php-ads-sdk for Marketing API?
This SDK is made for the Graph API

lpavlovi:
@yguedidi That is a good point, I don't know why facebook/facebook-php-ads-sdk was not used for this.
I think that @janhapke (the one who opened issue #774) would be the best to answer that.

janhapke:
For some use cases we are using the Graph SDK directly instead of going through the Ads SDK because that way we have better control over the amount of data we fetch in a single request (and by extension the total number of requests). These are cases where we are fetching 100s or 1000s of objects (e.g. campaigns).

yguedidi:
Looks like a good use case, @SammyK @tolbon what do you think?
@janhapke could you confirm that there is no other `stop_time` in the Graph that is not a datetime? Thanks

SammyK:
It's a BC change but since it's in master, I'm OK with it. :)

yguedidi:
@janhapke I'll merge this, but just for reference. You should not rely on the master version for you project, master is the development branch and it may change a lot before a release is tagged.

janhapke:
> @janhapke could you confirm that there is no other stop_time in the Graph that is not a datetime?

I'm afraid I do not know enough about the Graph to be able to confirm this.

> @janhapke I'll merge this, but just for reference. You should not rely on the master version for you project, master is the development branch and it may change a lot before a release is tagged.

This is fine by me. I can handle both \DateTime and string, so I am not depending on this change.

tolbon:
I think stop_time is referenced only here 
https://developers.facebook.com/docs/marketing-api/ad-rules-getting-started/v2.9

I search directly with search bar and I follow all first node routes proposed by graph api docs
https://developers.facebook.com/docs/graph-api/reference/
and not find any other stop_time in graph api.

yguedidi:
Thank you @tolbon!!

PR #786, opened 2017-05-20
fix mcrypt_create_iv

melbahja:
mcrypt_create_iv was DEPRECATED in php7+


SammyK:
Thanks for the PR @melbahja! But this check is already [implemented in the factory](https://github.com/facebook/php-graph-sdk/blob/5.5/src/Facebook/PseudoRandomString/PseudoRandomStringGeneratorFactory.php#L80-L83).

melbahja:
Awwh, i did not read the file name, it seems I did not sleep very well last night :)

PR #801, opened 2017-06-26
Master

mvcmaker:
Try to avoid to pass the appsecret_proof request var when is not necessary to be passed, due this request var produces error when is not mandatory to be passed.


yguedidi:
@mvcmaker Thanks for you first PR, we are honored! Your targeted the right branch (5.5), but looks like you based you work on the master one (that's why there is all thoses commit in the diff). Please base you local work on 5.5 anr push again

SammyK:
Hey @mvcmaker! The `appsecret_proof` is enabled by default as a security measure. If the app is not configured to require an `appsecret_proof`, a valid `appsecret_proof` will still be accepted. Getting an invalid `appsecret_proof` error means the access token you're attempting to use is not the same one that is owned by the app.

mvcmaker:
Hi @yguedidi and @SammyK what do you mean with base my local work on 5.5 and push again? This means that you want to put this pull request into version 5.5? I'm going to do it now. Thanks, but writing better the purposes, the use of appsecret_proof should be optional, to be configured by a method, by example $fb->disableAppSecretProof()

yguedidi:
@mvcmaker you should read this: https://opensource.guide/how-to-contribute/
Especially this part: https://opensource.guide/how-to-contribute/#opening-a-pull-request
Hope it will help you.

To quickly answer you, basicly it's:
- configure the `upstream` remote
- `git fetch upstream`
- `git checkout -b my_branch_name upstream/5.5`
- make changes and commit
- `git push -u origin my_branch_name`
- open the PR and target 5.5
- then the diff will show only changes you made above 5.5

But you really should read the github doc! ;)

yguedidi:
@mvcmaker and about the appsecret_proof, as it's about security, and because security is an important thing, we won't allow to disable it.

PR #803, opened 2017-06-27
Master

mvcmaker:
Try to avoid to pass the appsecret_proof request var when is not necessary to be passed, due this request var produces error when is not mandatory to be passed. You can by example add a method called "disableAppSecretProofSending()" in the FacebookFacebook class that avoids the use of appsecret_proof sending.


yguedidi:
@mvcmaker as we won't allow to disable the appsecret_proof, I'm closing this one. Thanks!

PR #806, opened 2017-06-29
Remove PHP5.6 support

Nyholm:

PR #809, opened 2017-06-30
Lets go to php7.1

tolbon:
PHP 7.1


yguedidi:
@tolbon Thanks you for this big PR! But we don't know if we move to 7.0 or 7.1, let's wait for @SammyK feedback, I'm for 7.1 of course!
BTW, things should be done incrementally, first require PHP7 (like in #806), then cleanup the code base, and final add PHP7 features like scalar type hint. :)

tolbon:
No problem
Thanks :)

SammyK:
Hey folks! Sorry for not being able to be quick with my responses. I've been trying to wrap up a huge client project the past month. :)

Considering [active support for PHP 7.0 ends in just 5 months](http://php.net/supported-versions.php) and the next version of this SDK probably won't come up until at least then, I'm down for requiring a minimum of PHP 7.1. :)

SammyK:
I forgot to mention that I'm not a fan of adding `declare(strict_types=1);` everywhere though. What's the benefit of adding this to the Facebook PHP SDK?

tolbon:
> I forgot to mention that I'm not a fan of adding declare(strict_types=1); everywhere though. What's the benefit of adding this to the Facebook PHP SDK?

@SammyK, @yguedidi 
Yeah I see what do you mean, I use it in my work and PHP Storm alert me about that.
I think it's not a important benefit in some case cause PHP make implicit cast. 
So it's better for retro-compatibility 

if you do this 
```php
function add(int $a, int $b)
{
    return $a + $b
}

add(1, 2) // = 3
add('1', "2") // implicit convert of 1 and 2 to int = 3
add(1, "Hello") // CRASH cause "Hello" isn't possible to convert in int
```

SammyK:
Yeah, I don't see a benefit for going strict types for the PHP SDK. :)

PR #811, opened 2017-06-30
Fix PHPDoc wrong type

tolbon:
Fix wrong type. Because Facebook didn't transmit int anymore string is for better compatibility

### DIFF
--- a/src/Facebook/Facebook.php
+++ b/src/Facebook/Facebook.php
@@ -576,7 +576,7 @@ class Facebook
     /**
      * Upload a video in chunks.
      *
-     * @param int $target The id of the target node before the /videos edge.
+     * @param string $target The id of the target node before the /videos edge.
      * @param string $pathToFile The full path to the file.
      * @param array $metadata The metadata associated with the video file.
      * @param string|null $accessToken The access token.
--- a/src/Facebook/FacebookBatchResponse.php
+++ b/src/Facebook/FacebookBatchResponse.php
@@ -92,7 +92,7 @@ class FacebookBatchResponse extends FacebookResponse implements IteratorAggregat
     /**
      * Add a response to the list.
      *
-     * @param int        $key
+     * @param mixed      $key
      * @param array|null $response
      */
     public function addResponse($key, $response)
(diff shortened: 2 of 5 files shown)


yguedidi:
@tolbon should PHPDoc fixes goes in this branch? @SammyK what do you think?

yguedidi:
@tolbon it's been a while... How are you? :)
To be sure this can be merged in 5.x, could you update it so that instead of changing types, just add the right one next to the existing one, and create a PR targeting master that make the right changes?
Thanks

src/Facebook/FileUpload/FacebookTransferChunk.php
```diff
@@ -106,7 +106,7 @@ public function getUploadSessionId()
     /**
      * Check whether is the last chunk
      *
-     * @return bool
+     * @return boolean
```
yguedidi: I prefer bool personally

src/Facebook/Facebook.php
```diff
@@ -576,7 +576,7 @@ public function videoToUpload($pathToFile)
     /**
      * Upload a video in chunks.
      *
-     * @param int $target The id of the target node before the /videos edge.
+     * @param string $target The id of the target node before the /videos edge.
```
SammyK: Since these changes are targeting the `5.x` branch, let's go with union type syntax for backward compatibility. So in this case, it'd be `int|string`. :)

src/Facebook/FacebookBatchResponse.php
```diff
@@ -92,7 +92,7 @@ public function setResponses(array $responses)
     /**
      * Add a response to the list.
      *
-     * @param int        $key
+     * @param mixed      $key
```
SammyK: Can we just do `int|string` here too? Why `mixed`? Would `$key` be anything else other than int or string? :)

src/Facebook/FileUpload/FacebookTransferChunk.php
SammyK: I'm with @yguedidi on this one - `bool` is the type declaration keyword in PHP 7, so let's go with that one. :)

PR #813, opened 2017-07-05
PHP 7.1

### LINKED ISSUE #787: [6.0] Require PHP 7.0 or greater
I know we've discussed supporting PHP 5.6 in the next version of the PHP SDK, but since PHP 5.6 is in ["security fixes only"](http://php.net/supported-versions.php) mode, we should consider requiring PHP 7.0 in version 6 of the PHP SDK.

*But wait! Lots of people are still on 5.6!* Yes, which is why we should put v5 of the PHP SDK in the same support mode as PHP 5.6 (security fixes only). We could maintain v5 with security support until PHP 5.6 is EOL on *Dec 31st, 2018*.

With requiring PHP 7.0 and greater we can take advantage of scalar type hints & return values thus giving us the flexibility to continually improve the SDK with static analysis tools. This requirement will also allow us to drop the CSPRNG dependency entirely.

There seems to be a movement of widely-used packages starting to require PHP 7 for their latest versions while maintaining support for PHP 5.6 until it's EOL. This is a great change in pace from previous PHP upgrades. Let's join them in this important movement! :)

yguedidi:
Closes #787

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: php
 
 php:
-  - 5.6
   - 7.0
   - 7.1
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,5 @@
 language: php
 
-php:
-  - 7.0
-  - 7.1
-
 sudo: false
 
 cache:
@@ -12,6 +8,8 @@ cache:
 
 matrix:
   include:
+    - php: 7.0
+    - php: 7.1
     - php: hhvm
       dist: trusty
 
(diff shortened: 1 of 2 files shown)


yguedidi:
@SammyK I'm trying to make the hhvm build green :)

SammyK:
Oh crap - did I mess you up by merging? :)

yguedidi:
Just a bit, but not a problem! ;)

SammyK:
Sorry! I got excited. Lol. I'll be hands-off unless you give me the green line. :)

PR #814, opened 2017-07-05
Add HHVM build but allow to fail

yguedidi:


yguedidi:
@SammyK you can merge this as is, I add HHVM builds in a PHP7 mode, but I can't manage to make them work as there are issue with HHVM and composer fixed in HHVM 3.20.2 (see [here](http://hhvm.com/blog/2017/06/07/hhvm-3-20.html))

PR #817, opened 2017-07-06
Fix loose of information when GraphNode convertion

### LINKED ISSUE #700: Only "data" key is cast to GraphNode
I'm retrieving a page's ratings including their id by requesting: `/v2.8/*PAGE_ID*/ratings?fields=open_graph_story`. The response has the following format:

```json
{
  "data": [
    {
      "open_graph_story": {
        "message": "*REVIEW_TEXT*",
        "start_time": "2016-11-08T14:09:42+0000",
        "type": "places.rates",
        "data": {
          "rating": {
            "value": 5,
            "scale": 5
          },
          "review_text": "*REVIEW_TEXT*",
          "language": "en",
          "generic_place": {
            "id": "*PAGE_ID*",
            "title": "*PAGE_NAME*",
            "type": "place",
            "url": "*PAGE_URL*"
          }
        },
        "id": "*REVIEW_ID*"
      }
    },
  ]
}
```

Each object in the `data` array has a key `id` that contains the review's story id. However, it is dropped when the reviews are casted into `GraphNode` objects:

```
object(Facebook\GraphNodes\GraphEdge)#1878 (5) {
  ["items":protected]=>
  array(25) {
    [0]=>
    object(Facebook\GraphNodes\GraphNode)#1761 (1) {
      ["items":protected]=>
      array(1) {
        ["open_graph_story"]=>
        object(Facebook\GraphNodes\GraphNode)#1762 (1) {
          ["items":protected]=>
          array(4) {
            ["rating"]=>
            object(Facebook\GraphNodes\GraphNode)#1774 (1) {
              ["items":protected]=>
              array(2) {
                ["value"]=>
                int(5)
                ["scale"]=>
                int(5)
              }
            }
            ["review_text"]=>
            string(928) "*REVIEW_TEXT*"
            ["language"]=>
            string(2) "en"
            ["generic_place"]=>
            object(Facebook\GraphNodes\GraphNode)#1760 (1) {
              ["items":protected]=>
              array(4) {
                ["id"]=>
                string(15) "*PAGE_ID*"
                ["title"]=>
                string(10) "*PAGE_NAME*"
                ["type"]=>
                string(5) "place"
                ["url"]=>
                string(35) "*PAGE_URL*"
              }
            }
          }
        }
      }
    }
  }
}
```

Responsible is this:

```php
// Sometimes Graph is a weirdo and returns a GraphNode under the "data" key
$data = $data['data'];
```

Graph is being a weirdo and doesn't include all information under the `data` key.

tolbon:
fixes #700
Maybe a patch to this issue #700

### DIFF
--- a/src/Facebook/GraphNodes/GraphNodeFactory.php
+++ b/src/Facebook/GraphNodes/GraphNodeFactory.php
@@ -303,7 +303,7 @@ class GraphNodeFactory
             if (static::isCastableAsGraphEdge($data['data'])) {
                 return $this->safelyMakeGraphEdge($data, $subclassName, $parentKey, $parentNodeId);
             }
-            if (count($data) === 1){
+            if (count($data) === 1) {
             // Sometimes Graph is a weirdo and returns a GraphNode under the "data" key
                 $data = $data['data'];
             }


tolbon:
I miss something wait please

tolbon:
That's good now

yguedidi:
Not sure about this solution... I'll investigate when I have some time.

tolbon:
I am just thinking about that but I need to test nested request. 
What do you think for better solution ? Merge all extra variable in ['data'] ?
The only problem with my solution is : you have graphNode with extra variable and an array 'data' with the rest of variable 
And you do 
´´´php
getField('data')['name of field in data']
´´´

jonny-no1:
This pull request now says
> @tolbon wants to merge 3 commits into `facebook:master` from `unknown repository`.

I've googled it and it sounds like the branch was deleted on the fork. There also hasn't been a reply from them in over a year. I'll open a duplicate PR with @tolbon's fix and my test from #1034.

tolbon:
@foaly-nr1 If you can do this very helpfull, thanks.

jonny-no1:
I've created a duplicate of this PR under #1085. The tests fail so this is not a fix for #700.

```
There was 1 failure:

1) Facebook\Tests\GraphNode\GraphNodeFactoryTest::testAGraphNodeWithARootDataKeyWillConserveId
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
     'id' => '123'
-    'name' => 'Foo McBar'
-    'link' => 'http://facebook/foo'
+    'data' => Array (...)
```

I suggest to reconsider #1034 which offers a solution.

tolbon:
ok I remember test with that 

```json
{
  "id": "123456",
  "picture": {
    "data": {
      "is_silhouette": false,
      "url": "https://scontent.xx.fbcdn.net/p50x50/foo.jpg"
    }
  }
}
```
yours is certainly better and resolve more case :+1:

yguedidi:
@foaly-nr1 thanks! How I see the fix, to be generic and don't destructure the Graph response:
- if only a `data` key and no other keys (like `picture` in @tolbon above example): use the data for the node (`$data = $data['data']`)
- otherwise, no changes, `data` key is a normal field like any other

yguedidi:
@foaly-nr1 @tolbon what do you think about this? :)

tolbon:
It's ok for me

@foaly-nr1 do you have true facebook call where your test is like that ?

jonny-no1:
@yguedidi yes that would work as it would preserve an `id` element that is next to a `data` element. I do believe that is the right way forward as it minimises data modification. Quite a big backwards compatibility break though?

I can update https://github.com/facebook/php-graph-sdk/pull/1085 accordingly.

yguedidi:
@foaly-nr1 good catch! It is a BC break.. I'll think about a way to fix the bug in a BC way when I have time (in fact I already have an idea that I need to confirm, but in France it's time to sleep). Meanwhile, any proposal to achieve that result? :)

yguedidi:
~~@foaly-nr1 my solution is a totally different solution from the proposed one, is it OK for you if I open my own PR?~~
EDIT: OK, basicly my solution is the same than the one you proposed in #1034, but extending it to more than just the `id` key. I'll continue the discussion in the other PR.

PR #819, opened 2017-07-06
Use latest PHPUnit and Prophecy

yguedidi:


yguedidi:
@SammyK maybe it's easier to review commit by commit :)

PR #825, opened 2017-07-14
Fix phpdoc

h4kuna:


yguedidi:
Thank you @h4kuna for your first contribution!!

PR #844, opened 2017-08-16
Fix phpdoc @return instead of @returns

### LINKED ISSUE #757: Using @returns instead of @return causes trouble
Hello,

I am getting the following error from using your SDK version 5.4.4 in a project with doctrine (shortened):

```
  Uncaught Exception: Doctrine\Common\Annotations\AnnotationException
  
  Message
    [Semantical Error] The annotation "@returns" in method
    Facebook\SignedRequest::split() was never imported. Did you maybe
  forget
    to add a "use" statement for this annotation?

  […]
```

The error can be easily fixed by using @return instead of @returns, [as proposed by phpdoc](https://phpdoc.org/docs/latest/references/phpdoc/tags/return.html). I guess you just mixed up with jsdoc :wink:

In particular, the problem occures at the following positions:

`src/Facebook/SignedRequest.php`:
- [L167](https://github.com/facebook/php-graph-sdk/blob/5.4/src/Facebook/SignedRequest.php#L167)
- [L185](https://github.com/facebook/php-graph-sdk/blob/5.4/src/Facebook/SignedRequest.php#L185)
- [L205](https://github.com/facebook/php-graph-sdk/blob/5.4/src/Facebook/SignedRequest.php#L205)

Please replace those.

gnat42:
PhpDoc expects @return, @returns was being used instead

This fixes #757

### DIFF
--- a/src/Facebook/SignedRequest.php
+++ b/src/Facebook/SignedRequest.php
@@ -164,7 +164,7 @@ class SignedRequest
     /**
      * Splits a raw signed request into signature and payload.
      *
-     * @returns array
+     * @return array
      *
      * @throws FacebookSDKException
      */
@@ -182,7 +182,7 @@ class SignedRequest
      *
      * @param string $encodedSig
      *
-     * @returns string
+     * @return string
      *
      * @throws FacebookSDKException
      */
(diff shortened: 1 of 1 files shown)


SammyK:
Thanks so much @gnat42! :)

PR #854, opened 2017-09-20
Added Hack Compatibility

### LINKED ISSUE #853: Hack Incompatibility
The current 5.x and master branches are incompatible with Hack.  Within release 5.6.1 the incompatibility stems from two distinct issues:

* The Collection class had a namespace conflict with Hack's `HH\Collection` (https://docs.hhvm.com/hack/reference/interface/HH.Collection/).

* The `parse_str()` PHP function is disabled by default in Hack due to security concerns.  Furthermore, PHP is deprecating the function in PHP 7.2+.

Is there interest in maintaining compatibility with Hack and PHP within this project, or would it be preferable for the Hack fork to be published on Packagist separately.

justinwray:
* The Collection class had a namespace conflict with Hack's `HH\Collection` (https://docs.hhvm.com/hack/reference/interface/HH.Collection/).  As a result, the Collection class has been renamed to GraphCollection.

* The `parse_str()` PHP function is disabled by default in Hack due to security concerns.  Furthermore, PHP is deprecating the function in PHP 7.2+.  As a result, the usage of `parse_str()` has been replaced with a `Util` class and `parseUrlQuery()` method.  This solution was pulled from [facebook/facebook-php-ads-sdk](https://github.com/facebook/facebook-php-ads-sdk) ([FacebookAds/Http/Util.php](https://github.com/facebook/facebook-php-ads-sdk/blob/master/src/FacebookAds/Http/Util.php)).

* Fixes #853 - For the 5.x branch

### DIFF
--- a/src/Facebook/Http/Util.php
+++ b/src/Facebook/Http/Util.php
@@ -24,7 +24,8 @@
 
 namespace Facebook\Http;
 
-abstract class Util {
+abstract class Util
+{
 
   /**
    * Avoid parse_str() for HHVM compatibility
@@ -35,14 +36,15 @@ abstract class Util {
    * @param $query_string
    * @return array
    */
-  public static function parseUrlQuery($query_string) {
-    $query = array();
-    $pairs = explode('&', $query_string);
-    foreach ($pairs as $pair) {
-      list($key, $value) = explode('=', $pair);
-      $query[$key] = urldecode($value);
-    }
+    public static function parseUrlQuery($query_string)
+    {
+        $query = array();
+        $pairs = explode('&', $query_string);
+        foreach ($pairs as $pair) {
+            list($key, $value) = explode('=', $pair);
+            $query[$key] = urldecode($value);
+        }
 
-    return $query;
-  }
+        return $query;
+    }
 }
--- a/src/Facebook/Http/Util.php
+++ b/src/Facebook/Http/Util.php
@@ -1,0 +1,48 @@
+<?php
+/**
+ * Copyright (c) 2014-present, Facebook, Inc. All rights reserved.
+ *
+ * You are hereby granted a non-exclusive, worldwide, royalty-free license to
+ * use, copy, modify, and distribute this software in source code or binary
+ * form for use in connection with the web services and APIs provided by
+ * Facebook.
+ *
+ * As with any software that integrates with the Facebook platform, your use
+ * of this software is subject to the Facebook Developer Principles and
+ * Policies [http://developers.facebook.com/policy/]. This copyright notice
+ * shall be included in all copies or substantial portions of the software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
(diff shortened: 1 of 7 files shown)


SammyK:
Thanks for the PR @justinwray! 

1. Changing the name of the `Collection` class is a breaking change so we can't change this in the 5.x branch. Even though the class name is the same as `HH\Collection`, they are in different namespaces. So you could just do this in your scripts:

```php
use Facebook\ GraphNodes\Collection as GraphCollection;
```

2. I'm thinking we move the `parse_str()` polyfill to the [polyfills.php](https://github.com/facebook/php-graph-sdk/blob/5.x/src/Facebook/polyfills.php) file instead of creating a `Util` class. PHP 7.2 isn't deprecating `parse_str()`, it's just deprecating the omission of the second argument to dump all the vars into the global namespace (see [the docs](http://php.net/parse_str#refsect1-function.parse-str-parameters)). The Facebook PHP SDK makes use of the second argument so we're don't going to have to worry about any deprecation messages. :)

@yguedidi Do you have anything thoughts on this as well? :)

justinwray:
@SammyK Not a problem at all!

Adding `use Facebook\GraphNodes\GraphNode\Collection as GraphCollection;` to our class solved the conflict with the `Collection` class.  I will drop that from the PR, though I do think it might be wise to document this (either in the code) or within the php-graph-sdk docs, as it will be a common issue for anyone using HHVM.

I will also work on moving the `parse_str()` replacement to `polyfills.php`.

yguedidi:
@SammyK I agree with should support HHVM, even in 5.5 as we don't know when 6.0 will be released :)

src/Facebook/Http/Util.php
```diff
@@ -0,0 +1,50 @@
...
+ * THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ *
+ */
+
+namespace Facebook\Http;
+
+abstract class Util
```
yguedidi: looks like file is a duplicate, and not used

src/Facebook/Url/Util.php
```diff
@@ -0,0 +1,50 @@
...
+  /**
+   * Avoid parse_str() for HHVM compatibility
+   * This implementation is not a complete sobstitute, but covers all the
+   * requirements of the Facebook Url namesapce.
+   *
+   * @see hhvm.hack.disallow_dynamic_var_env_funcs
+   * @param $query_string
+   * @return array
+   */
+    public static function parseUrlQuery($query_string)
```
yguedidi: instead of a static function in that utility class, I'd move the function definition in the `FacebookUrlManipulator` (as it's used only there, rename it to `parse_str`, and wrap it with a `if (!function_exists('parse_str'))`.

yguedidi:
@SammyK do we really need to support Hack..?

SammyK:
@yguedidi I think ideally the Facebook PHP SDK would support HHVM as it's one of the official Facebook products. But since we're the only two maintainers and we don't use HHVM, it's kinda hard to maintain HHVM support on our volunteer time. I say if Facebook really wants HHVM support for the PHP SDK, they can provide a maintainer to maintain HHVM support. :)

@justinwray Thanks for this PR! Since this has a lot of different issues associated with it, I'm going to go ahead and close this. If you want to submit a smaller PR with some of the specific things we've been talking about, please feel free and we'll keep the conversation going! :)

yguedidi:
@SammyK totally agree!

PR #883, opened 2017-11-08
Added possibility to choose the login display type

jbereziecki:
I find it necessary to be able to choose the Facebook Login display type.

### DIFF
--- a/docs/examples/facebook_login.md
+++ b/docs/examples/facebook_login.md
@@ -99,3 +99,14 @@ $_SESSION['fb_access_token'] = (string) $accessToken;
 // You can redirect them to a members-only page.
 //header('Location: https://example.com/members.php');
 ```
+
+## Login display type
+
+You might want to change the default 'page' display type. There is an enumeration that can be used to select the desired display when calling FacebookRedirectLoginHelper::getLoginUrl() method:
+
+* **async:** FacebookDisplayTypeHelper::DISPLAY_TYPE_ASYNC
+* **iframe:** FacebookDisplayTypeHelper::DISPLAY_TYPE_IFRAME
+* **page:** FacebookDisplayTypeHelper::DISPLAY_TYPE_PAGE
+* **popup:** FacebookDisplayTypeHelper::DISPLAY_TYPE_POPUP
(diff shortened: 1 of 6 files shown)


yguedidi:
@jbereziecki thank you for this PR!! And sorry for the delay..
Wahou! Complete PR with code, tests and doc!
Unfortunatly, according the [Facebook doc](https://developers.facebook.com/docs/facebook-login/manually-build-a-login-flow/) looks like the `display` parameters in not supported anymore, so I'm closing this.
Feel free to contribute to the SDK!

jbereziecki:
@yguedidi 
Thanks for finally looking into this.
Although not documented, the display type parameter is active when invoking the OAuth dialog under https://www.facebook.com/v2.10/dialog/oauth

I have just tested it and the different display types result in different dialogs being rendered. I need this a lot in my application.

yguedidi:
@jbereziecki how you test it? I just tested it using the Graph API Explorer, and here are what I saw:

- `async` and `iframe` don't work, I get an error
- `touch` and `wap` looks like exactly like the `page` display
- `popup` is working and has a different look than `page`

Based on that, only `page` and `popup` are useful.

So then, why would you like to use the `popup` display while it was made for the JS SDK I guess :)
What's your use case?

jbereziecki:
@yguedidi I tested it manually invoking the OAuth dialog, and for me it seems, that async returns 500 indeed, iframe requires public_profile scope, the rest of them is exactly as you described.

I need to be able to select the display type because we are invoking the login dialog from different clients - it might be either a desktop app (the page display is fine here) or a popup in a web app (we need the popup display here to strip down the FB footer etc.).

I might limit the display type enumeration to only contain page and popup if it's the case.

yguedidi:
@jbereziecki just to understand better, why don't you use the Javascript SDK to login using a popup?
Except if you app already live inside a popup, but then, it's really a specific usage. So I think we won't support it in the SDK.
Why don't you extend the `FacebookRedirectLoginHelper`? :)

jbereziecki:
@yguedidi Javascript SDK is not suitable for the desktop client and duplicating the logic is a bad idea.

I'll just extend the SDK locally, wanted to avoid it tho due to possible later incompatibilities. Thanks for your support.

Issue #857, opened 2017-09-28
Incorrect phpdoc return type for GraphUser->getBirthday()


julienbornstein:
Hello 😀
It seems that there's an incorrect return type in phpdoc In src/Facebook/GraphNodes/GraphUser.php for method getBirthday()
Current : @return \DateTime|null
Should be : @return Facebook\GraphNodes\Birthday|null

yguedidi:
@julienbornstein Thanks for the notice! Can you send a PR that fix it? :)

yguedidi:
Closed by #885

PR #885, opened 2017-11-21
Fix PHPDoc return type #857

julienbornstein:
getBirthday () return Birthday type


yguedidi:
Consider this a bugfix, so accepted! Thank you @julienbornstein for your first contribution!

PR #891, opened 2017-12-03
cURL follow fix

### LINKED ISSUE #886: not following 30X redirects
https://github.com/facebook/php-graph-sdk/blob/2f9639c15ae043911f40ffe44080b32bac2c5280/src/Facebook/HttpClients/FacebookCurlHttpClient.php#L101

curl option **CURLOPT_RETURNTRANSFER** is not about redirects
there should be new default option **CURLOPT_FOLLOWLOCATION** with value **true**

ryanml:
Closes #886

Currently in the openConnection method, the comments imply that 301 redirects are followed with the CURLOPT_RETURNTRANSFER option. This is incorrect, as that option enables a cURL response to be returned as a string, rather than being written as output. 

Added to the options with this commit is CURLOPT_FOLLOWLOCATION, which will enable cURL to follow 301 redirects.


ryanml:
@yguedidi thanks for the review, I've removed the FOLLOWLOCATION directives, so the only change with this is now the comment description for RETURNTRANSFER

yguedidi:
Thank you @ryanml! Congrats for your first PR here!

ryanml:
@yguedidi thanks so much! I look forward to contributing more.

PR #899, opened 2017-12-06
Remove Collection class

### LINKED ISSUE #440: [6.0][Proposal] Remove Collection class

yguedidi:
closes #440 

This only duplicates the Collection class code in GraphNode and GraphEdge.
A rework of the API of thoses two classes will come after in separate PRs.


yguedidi:
Thank you @Nyholm! I don't like duplication neither! My goal is to scope PRs, so that they don't have to much resposabilities/changes ;)

PR #900, opened 2017-12-06
Refactoring tests

carusogabriel:
I've refactored some tests, using:
- `assertContains` instead of `in_array` function;
- `assertCount` instead of `count` function;
- `assertInternalType`, avoiding using `is_*` functions;
- `assertArrayNotHasKey` instead of `isset` function;


yguedidi:
Thanks you @carusogabriel!! Congrats for your first PR here ;)

PR #905, opened 2017-12-14
Update FacebookRedirectLoginHelper.php

### LINKED ISSUE #877: Facebook SDK not working for me (solved) (bug)
Just a quick note to explain the problem and its solution: copying and pasting example code returns an error 191 telling that the callback url isn't allowed and pointing you to app settings inside facebook.
What? It's impossible! Check multiple times.. No.. the url is right! So, why?

Solution 1 add callback url to getAccessToken in your code:
$accessToken = $helper->getAccessToken("http://www.yoursite.com/fb-callback.php");

Solution 2 edit facebook sdk code:
https://github.com/facebook/php-graph-sdk/blob/2f9639c15ae043911f40ffe44080b32bac2c5280/src/Facebook/Helpers/FacebookRedirectLoginHelper.php#L226
Edit this line, find ['state'] and add code ['state','code']
Code param isn't stripped from callback url by the helper so it's sent to fb and you get the error.

Miles8of9:
Closes #877

Solves bug explained in issue https://github.com/facebook/php-graph-sdk/issues/877 

Thanks for your attention.

### DIFF
--- a/src/Facebook/Helpers/FacebookRedirectLoginHelper.php
+++ b/src/Facebook/Helpers/FacebookRedirectLoginHelper.php
@@ -223,7 +223,7 @@ class FacebookRedirectLoginHelper
 
         $redirectUrl = $redirectUrl ?: $this->urlDetectionHandler->getCurrentUrl();
         // At minimum we need to remove the state param
-        $redirectUrl = FacebookUrlManipulator::removeParamsFromUrl($redirectUrl, ['state']);
+        $redirectUrl = FacebookUrlManipulator::removeParamsFromUrl($redirectUrl, ['state','code']);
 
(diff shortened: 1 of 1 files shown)


src/Facebook/Helpers/FacebookRedirectLoginHelper.php
```diff
@@ -223,7 +223,7 @@ public function getAccessToken($redirectUrl = null)
 
         $redirectUrl = $redirectUrl ?: $this->urlDetectionHandler->getCurrentUrl();
         // At minimum we need to remove the state param
```
yguedidi: Can you please update this comment?

yguedidi:
Closing in favor of #913, thanks @Miles8of9

PR #906, opened 2017-12-18
Rework GraphNode API

yguedidi:
API design decision:
- A node **is not** a "collection of fields", it doesn't make sense to loop over them, count them, etc so all list behavior functions are removed. We care about accessing each specific field.
- We have the `getField()` function, so array access is not needed and less semantic.
As we want to provide as much typed node classes as possible with semantic getter for each field, the array access becomes even more useless.
Not having it will force use to maintain an updated types node API, which is goot IMO.
- Utility functions than were public to make their testing easier are now private, they are not part of what a graph node is.

Better review commit by commit :)

### DIFF
--- a/src/GraphNode/GraphNode.php
+++ b/src/GraphNode/GraphNode.php
@@ -25,7 +25,7 @@ namespace Facebook\GraphNode;
 /**
  * @package Facebook
  */
-class GraphNode implements \ArrayAccess, \Countable, \IteratorAggregate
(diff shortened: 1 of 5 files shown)


yguedidi:
@Nyholm here the first API rework ;)

yguedidi:
Thanks @SammyK ! :)
Next step: GraphEdge, I'll work on them soon, I already have a WIP since that time haha

PR #913, opened 2018-01-09
Strip 'code' param

### LINKED ISSUE #877: Facebook SDK not working for me (solved) (bug)

martinstuecklschwaiger:
This pull request fixes #877, since there hasn't been any activity in #905 for 25 days. 

I also adjusted the test to reflect the change and prevent accidental breaking in the future. 

I also changed the return value of `FooRedirectLoginOAuth2Client::getAccessTokenFromCode()` since that's the specified behavior in `OAuth2Client`. 

Please note that this fix is needed for “Strict Redirect URI Matching” which will be required by Facebook in March.

### DIFF
--- a/tests/Helpers/FacebookRedirectLoginHelperTest.php
+++ b/tests/Helpers/FacebookRedirectLoginHelperTest.php
@@ -44,6 +44,9 @@ class FacebookRedirectLoginHelperTest extends \PHPUnit_Framework_TestCase
     protected $redirectLoginHelper;
 
     const REDIRECT_URL = 'http://invalid.zzz';
+    const FOO_CODE = "foo_code";
+    const FOO_STATE = "foo_state";
+    const FOO_PARAM = "some_param=blah";
 
     protected function setUp()
     {
@@ -94,10 +97,10 @@ class FacebookRedirectLoginHelperTest extends \PHPUnit_Framework_TestCase
     public function testAnAccessTokenCanBeObtainedFromRedirect()
     {
         $this->persistentDataHandler->set('state', 'foo_state');
-        $_GET['state'] = 'foo_state';
-        $_GET['code'] = 'foo_code';
+        $_GET['state'] = static::FOO_STATE;
+        $_GET['code'] = static::FOO_CODE;
 
-        $fullUrl = self::REDIRECT_URL . '?state=foo_state&code=foo_code&some_param=blah';
+        $fullUrl = self::REDIRECT_URL . '?state=' . static::FOO_STATE . '&code=' . static::FOO_CODE . '&' . static::FOO_PARAM;
 
         $accessToken = $this->redirectLoginHelper->getAccessToken($fullUrl);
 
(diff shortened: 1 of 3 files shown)


Savageman:
Thanks a ton for this! I was just working on the upgrade for "Strict Redirect URI Matching" today and it's not working because of that...
We need this PR accepted :)

martinstuecklschwaiger:
@yguedidi I hope I'm not annoying you but is there any change to get this pull request (or #905) merged soon?

tests/Helpers/FacebookRedirectLoginHelperTest.php
```diff
@@ -97,9 +97,14 @@ public function testAnAccessTokenCanBeObtainedFromRedirect()
         $_GET['state'] = 'foo_state';
         $_GET['code'] = 'foo_code';
 
-        $accessToken = $this->redirectLoginHelper->getAccessToken(self::REDIRECT_URL);
+        $fullUrl = self::REDIRECT_URL . '?state=foo_state&code=foo_code&some_param=blah';
```
gmertk: May use const for the text "foo_state" "foo_code" and "some_param=blah"

gmertk:
@yguedidi if it looks good to you, let's get this going! :)

martinstuecklschwaiger:
@gmertk better now?

yguedidi:
@martinstuecklschwaiger 5.6.2 released! Thanks again for the PR! Thanks @gmertk for your help in review!

hao214:
my callback url is http://www.example.dk/fb/fbcallback.php?id=### how can I define this

please help

yguedidi:
@hao214 if you need help about using it, you should ask your questions on [StackOverflow](https://stackoverflow.com/) and tag them with the `facebook-php-sdk` tag.

alexwai:
Finally i manage to find the problem when using the  PHP SDK.

The message that always shows is "The domain of this URL isn't included in the app's domains."

On the callback page in the getAccessToken need to specify the call back url even thou it is optional

```
try 
{
	$accessToken = $helper->getAccessToken('https://place-your-callback-url-here');
} 
catch (\Facebook\Exceptions\FacebookResponseException $e) 
{
	echo $e->getMessage();
} 
catch (\Facebook\Exceptions\FacebookSDKException $e) 
{

	echo $e->getMessage();
}
if (!$accessToken) 
{
	echo 'FAIL';
}
```

Hope this helps......

Issue #916, opened 2018-01-12
 implement strict mode compatible authorization


kacurez:
**`UPDATE`** : if I generate the state parameter on my own then `response_type=code` auth is working with strict mode turned on. So implementing oauth via `response_type=token` is not necessary, moreover `response_type=token` is for client-side authorization as it generates callback url with url params separated by hash(`#`) and not question mark(`?`).  However I don't know why state value generated by this sdk causes the problem described below.

probably related to #873 
Facebook is enforcing strict mode to be used in the authorization process.
https://developers.facebook.com/docs/facebook-login/security/#strict_mode

![image](https://user-images.githubusercontent.com/1412120/34882627-76b4b914-f7b7-11e7-95e0-15904810ce6e.png)

I use this sdk to authorize a fb account and retrieve access token. However when I turn on strict mode, then it will fail with error message:
`Can't Load URL: The domain of this URL isn't included in the app's domains.`

I contacted fb support and been told not to use response_type=code. The `response_type=code` is hardwired in the authorization module of this sdk, see:
https://github.com/facebook/php-graph-sdk/blob/5.x/src/Facebook/Authentication/OAuth2Client.php#L140

They recommended me to use `response_type=token`, so I tried it(manually updating source code) and it worked. 
Could you please update the authorization module to comply strict mode policy?

sam4jips:
how did you set the state parameter manually? please teach me. I looked everywhere but couldn't find anywhere.

kacurez:
The api of this library does not offer to put state parameter so I ended up implementing the oauth process on my own without any use of this library.

sam4jips:
actually i found the solution. you can set state manually if you want.

$helper = $fb->getRedirectLoginHelper();
$helper->getPersistentDataHandler()->set('state', $state);
$helper->getPersistentDataHandler()->get('state');

claudeschneider:
We're getting the same problem - oAuth it fails with the same error mentioned above (domain of URL isn't included in the app's domain)

Facebook also gave the same response, that the presence of ?code= in the redirect URL materially changes the redirect URL, and thus stops strict mode from working.

Even though the official docs still mention using "code" to get a response that can be exchanged for an access token
https://developers.facebook.com/docs/facebook-login/manually-build-a-login-flow

If the PHP SDK hard-codes the "response_type=code" into the loginHelper URL, and that no longer works, it needs to be changed in the PHP-SDK.

cannond13:
[...]

PR #918, opened 2018-01-16
Allowing response_type to be overriden for authorization url retrieval

ryanml:
Currently the response type of the getAuthorizationUrl method of the OAuth2Client hard codes a response type of 'code'. Per #916, strict mode requires a response type of 'token'. With respects to backwards compatibility, this allows response type to be overridden with the 'token' literal for example, while defaulting to 'code'

### DIFF
--- a/src/Authentication/OAuth2Client.php
+++ b/src/Authentication/OAuth2Client.php
@@ -129,12 +129,12 @@ class OAuth2Client
      *
      * @return string
      */
-    public function getAuthorizationUrl($redirectUrl, $state, array $scope = [], array $params = [], $separator = '&')
+    public function getAuthorizationUrl($redirectUrl, $state, array $scope = [], array $params = [], $separator = '&', $response_type = 'code')
     {
         $params += [
             'client_id' => $this->app->getId(),
             'state' => $state,
-            'response_type' => 'code',
+            'response_type' => $response_type,
             'sdk' => 'php-sdk-' . Facebook::VERSION,
             'redirect_uri' => $redirectUrl,
             'scope' => implode(',', $scope)


ryanml:
cc: @yguedidi @kacurez

kacurez:
@ryanml parameter `$response_type` should be imho propagated from `getLoginUrl ` method in `FacebookRedirectLoginHelper`
https://github.com/ryanml/php-graph-sdk/blob/916-fix/src/Helper/RedirectLoginHelper.php#L123

at least this would work for me as I use it in the following pseudocode:

```php
$provider = new Facebook([
    'app_id'          => $this->appKey,
    'app_secret'      => $this->appSecret,
    'default_graph_version' => $this->getGraphApiVersion()
]);
$helper = $provider->getRedirectLoginHelper();
$loginUrl = $helper->getLoginUrl($callbackUrl, $permissions); // here should be $helper->getLoginUrl($callbackUrl, $permissions, '$',  'token')
return $this->redirect($loginUrl); // symfony redirect method
```
just to mention: If `response_type=token` is used then redirect callback method(that I implement) will receive short-lived access-token instead of code. So the redirect callback method should only check received state parameter, compare with the one stored previously in the session and also inspect received token via : https://developers.facebook.com/docs/facebook-login/manually-build-a-login-flow/#checktoken

kacurez:
I realised that `response_type=token` is for client-side authorization as it generates callback url with url params separated by hash(#) and not question mark(?). So implementing it on server would not help either. 
Finally I think that the problem is caused by state parameter value but don't know why. If I generate state "manually" then it works.

yguedidi:
Thanks @ryanml for the contribution!
Strict mode is now supported in version 5.6.2! So I'm closing this one.

PR #926, opened 2018-01-31
TODO v 6.0

m0uj:
Throw an InvalidArgumentException if "default_graph_version" is not set,
Remove method getProperty() since it is renamed to getField()


yguedidi:
@m0uj thanks for your contribution! But thoses changes are already part of master :)
I think you should fetch the latest master locally

PR #941, opened 2018-02-14
added post status update example

sohaibilyas:
a simple example to post status update on user's timeline.


yguedidi:
@sohaibilyas Thanks you for your pull request! This example looks like the post links one, so I'm not sure we should add that many examples in the code, especially when they are almost the same.
@SammyK what do you think?
(BTW, until we get an answer to this question, your PR needs to be rebased, and you should add a link to your example in `docs/README.md`)

SammyK:
@sohaibilyas Thanks so much for this PR! I agree with @yguedidi - this example is almost identical to the link example and would add clutter to the documentation in addition to the need for it to be maintained over time. I'd be open to refactoring the exiting links example to a status update example, but I'm thinking the link example is more useful in the majority of contexts that people use the SDK. Thanks again for the PR! :)

PR #944, opened 2018-02-16
Throw an InvalidArgumentException if "default_graph_version" is not set

ifaisalalam:
The SDK will now throw an InvalidArgumentException if default graph version is not set.


yguedidi:
Thank you @ifaisalalam for your first contribution!
I can't accept it, this change is a BC break (Backward Compatibility break), that's why the todo comment is about the next major version, v6.
Hope to see you contribute again to the SDK!

PR #953, opened 2018-02-28
Avoiding multiple return statements

ShubhamBansal1997:
Generally, as a programming point of multiple return statements are not considered good


yguedidi:
@ShubhamBansal1997 Thanks for the pull request, but I disagree with it.

In your code, first you add a level of indentation, which is basiclly something bad.
You also add the number of tests: if it's a string, `is_string` is called 2 times...
As this is mostly about opinion on code structure and brings ne benefits over the existing code, I'm closing this.

Hope to you continue contributing to the SDK!

PR #957, opened 2018-03-13
Adding `postJson` to support endpoints that expect 'application/json'.

yssk22:
Facebook Messenger POST graph endpoints receive 'application/json'
content type while the current SDK doesn't support that since it sends
the request body with multipart or url-encoded.

This patch introduces a new `postJosn()` method with the same
sigunature as `post()` and developers can call Messenger APIs with
a single SDK instead of developing their own API client.


yssk22:
close this since ['param' => json_encode(value)] can work with endpoints that show application/json examples.

PR #988, opened 2018-05-09
Updated default Graph API Version

JessJNielsen:
The default api version "v2.10" is deprecated and causes errors. 

It should be updated and we might as well move to 3.0 already.

### DIFF
--- a/src/Facebook/Facebook.php
+++ b/src/Facebook/Facebook.php
@@ -58,7 +58,7 @@ class Facebook
     /**
      * @const string Default Graph API version for requests.
      */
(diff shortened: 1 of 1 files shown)


yguedidi:
@JessJNielsen thanks for this contribution!
I sadly can't merge it as is: by targeting 5.x branch, this is a BC break, it will break all existing installations of the SDK.
But the right API version to updrage to is the lowest supported one by Facebook, as whatever the deprecated version you use, the Graph API will use the lowest one, so ne BC break.
The current one is 2.6, available until 2018-07-13.
2.10 is available until 2019-11-07, we can keep it as a default version.
So I'm closing this.

PR #1001, opened 2018-06-01
Fix resumable upload error "There was a problem uploading your video"

phaizullin:
Changed start_offset and end_offset from error code as written in documentation

### DIFF
--- a/src/Facebook/Exceptions/FacebookResponseException.php
+++ b/src/Facebook/Exceptions/FacebookResponseException.php
@@ -90,11 +90,22 @@ class FacebookResponseException extends FacebookSDKException
                 // Video upload resumable error
                 case 1363030:
                 case 1363019:
-                case 1363037:
                 case 1363033:
                 case 1363021:
                 case 1363041:
                     return new static($response, new FacebookResumableUploadException($message, $code));
+                case 1363037:
+                    $previousException = new FacebookResumableUploadException($message, $code);
+
+                    $startOffset = isset($data['error']['error_data']['start_offset']) ?
+                        $data['error']['error_data']['start_offset'] : null;
+                    $previousException->setStartOffset($startOffset);
+
+                    $endOffset = isset($data['error']['error_data']['end_offset']) ?
+                        $data['error']['error_data']['end_offset'] : null;
+                    $previousException->setEndOffset($endOffset);
+
+                    return new static($response, $previousException);
             }
         }
 
--- a/src/Facebook/Exceptions/FacebookResponseException.php
+++ b/src/Facebook/Exceptions/FacebookResponseException.php
@@ -97,12 +97,10 @@ class FacebookResponseException extends FacebookSDKException
                 case 1363037:
                     $previousException = new FacebookResumableUploadException($message, $code);
 
-                    $startOffset = isset($data['error']['error_data']['start_offset']) ?
-                        $data['error']['error_data']['start_offset'] : null;
+                    $startOffset = isset($data['error']['error_data']['start_offset']) ? (int) $data['error']['error_data']['start_offset'] : null;
                     $previousException->setStartOffset($startOffset);
 
-                    $endOffset = isset($data['error']['error_data']['end_offset']) ?
-                        $data['error']['error_data']['end_offset'] : null;
+                    $endOffset = isset($data['error']['error_data']['end_offset']) ? (int) $data['error']['error_data']['end_offset'] : null;
                     $previousException->setEndOffset($endOffset);
 
                     return new static($response, $previousException);
--- a/src/Facebook/Exceptions/FacebookResumableUploadException.php
+++ b/src/Facebook/Exceptions/FacebookResumableUploadException.php
@@ -30,4 +30,39 @@ namespace Facebook\Exceptions;
  */
 class FacebookResumableUploadException extends FacebookSDKException
 {
+    protected $startOffset;
+
+    protected $endOffset;
+
+    /**
+     * @return int
+     */
+    public function getStartOffset()
+    {
+        return $this->startOffset;
+    }
+
+    /**
+     * @param int $startOffset
+     */
+    public function setStartOffset($startOffset)
+    {
+        $this->startOffset = $startOffset;
+    }
+
+    /**
+     * @return int
+     */
+    public function getEndOffset()
+    {
+        return $this->endOffset;
+    }
+
+    /**
+     * @param int $endOffset
+     */
+    public function setEndOffset($endOffset)
+    {
+        $this->endOffset = $endOffset;
+    }
 }
--- a/src/Facebook/Exceptions/FacebookResumableUploadException.php
+++ b/src/Facebook/Exceptions/FacebookResumableUploadException.php
@@ -35,7 +35,7 @@ class FacebookResumableUploadException extends FacebookSDKException
     protected $endOffset;
 
     /**
-     * @return int
+     * @return int|null
      */
     public function getStartOffset()
     {
@@ -43,7 +43,7 @@ class FacebookResumableUploadException extends FacebookSDKException
     }
 
     /**
-     * @param int $startOffset
+     * @param int|null $startOffset
      */
     public function setStartOffset($startOffset)
     {
@@ -51,7 +51,7 @@ class FacebookResumableUploadException extends FacebookSDKException
     }
 
     /**
-     * @return int
+     * @return int|null
      */
     public function getEndOffset()
     {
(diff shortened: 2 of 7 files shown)


yguedidi:
@phaizullin Thanks for your contribution!
Why would you like include this behavior in the SDK? Especially that you can do the same in your code by allowing to throw and catching the exception

phaizullin:
@yguedidi Firstly, because it is mostly facebook internal error, it want to process another chunk (mostly random chunk) after processing current chunk, but uploadVideo method retry only current chunk and it returns same error 5 times via maxTriesTransfer method. 
I think that main function of sdk is minimize interactions between user code and api. If api wants to upload another chunk it must do it.
Secondly, if I catch the error, I will need to re-implement Facebook class methods

nasy:
I am also experiencing the same problems, it would be nice for this fix to get merged.

abhishekbh:
This is an important patch. What's the hold up in the merge or further comments?

SammyK:
This seems to be a common issue so I'd be +1 for merging. For the record, the [error message for `1363037`](https://developers.facebook.com/docs/graph-api/video-uploads#errors) is:

> There was a problem uploading your video. Please try uploading it again. error_data:{start_offset:0,end_offset:392695}

The recommendation is:

> Returned when the server receives an invalid start offset from the client. Try to parse the start_offset and end_offset from error message and do a retry again.

@phaizullin & @yguedidi Assuming we merge this in - we should probably figure out how this impacts the max retires. This PR technically adds a non-deterministic outcome as there is an edge case where Graph returns the `1363037` sub code every time. Wudya think? :)

src/Facebook/FileUpload/FacebookResumableUploader.php
```diff
@@ -121,6 +121,16 @@ public function transfer($endpoint, FacebookTransferChunk $chunk, $allowToThrow
                 throw $e;
             }
 
+            if ($e->getSubErrorCode() === 1363037) {
+                return new FacebookTransferChunk(
+                    $chunk->getFile(),
+                    $chunk->getUploadSessionId(),
+                    $chunk->getVideoId(),
+                    $e->getResponseData()['error']['error_data']['start_offset'],
```
SammyK: We should probably do some `isset()` checks on this and the next line just in case Graph changes the error data structure it responds with.

src/Facebook/FileUpload/FacebookResumableUploader.php
```diff
@@ -121,6 +121,16 @@ public function transfer($endpoint, FacebookTransferChunk $chunk, $allowToThrow
                 throw $e;
             }
 
+            if ($e->getSubErrorCode() === 1363037) {
```
SammyK: Let's make `1363037` a class constant. :)

yguedidi:
@phaizullin @SammyK I'm not that against this PR, I'm more against the implementation.

Let met explian: This solution looks more like a quick workaround to the issue.

How I conceptualy see it is more: you should catch a specific exception, that will give you access to a `getRetryableChunk()` or somethig like that, and you do what you want with it.

Enforcing an error handling internally looks bad to me. An SDK responsability is to provide the more detailed error, while handle it is user's code responsability: an SDK shouldn't has error no? :)

phaizullin:
@yguedidi @SammyK thanks for review. Please see the changes one more time)

abhishekbh:
@yguedidi what you're saying makes sense, however the current pattern the SDK sets up with the "super" Facebook class is that the client has a minimum amount of information about the upload process.

If the FB SDK expects the calling client to handle chunks and errors (not a bad idea), it should expose more methods to allow an efficient way of doing this. The only public method in the Facebook class at the moment is "uploadVideo".

In the meanwhile, is there another recommendation on working around this issue?

SammyK:
@abhishekbh You could work directly with the `Facebook\FileUpload\FacebookResumableUploader` class to better handle these cases, but it doesn't have retry logic built in since that's handled by the main `Facebook` service class. I'd love hear if @yguedidi has a better suggestion for handling this particular case though. :)

yguedidi:
How I see things:

- `FacebookResumableUploader` is the pure way, it does one try, if it fails, it throws
- `uploadVideo()` is a helper function, a convenient way to do a full upload in one run, with cool feature like auto-retry.

I see the proposed fix (return the retryable chunk), as one of those cool feature, so I think it should live in the `uploadVideo()` function and be counted in the number of tries. :)
So basicly, what annoys me here is that it's the `FacebookResumableUploader` who handle that special case.

phaizullin:
@yguedidi It is't entirely clear from your words what changes in the PR are needed.

src/Facebook/Exceptions/FacebookResponseException.php
```diff
@@ -90,11 +90,22 @@ public static function create(FacebookResponse $response)
...
-                case 1363037:
                 case 1363033:
                 case 1363021:
                 case 1363041:
                     return new static($response, new FacebookResumableUploadException($message, $code));
+                case 1363037:
+                    $previousException = new FacebookResumableUploadException($message, $code);
+
+                    $startOffset = isset($data['error']['error_data']['start_offset']) ?
+                        $data['error']['error_data']['start_offset'] : null;
```
yguedidi: please keep this in the same line. BTW, you should cast to int the offset

src/Facebook/Exceptions/FacebookResponseException.php
```diff
@@ -90,11 +90,22 @@ public static function create(FacebookResponse $response)
...
+                    $previousException->setStartOffset($startOffset);
+
+                    $endOffset = isset($data['error']['error_data']['end_offset']) ?
+                        $data['error']['error_data']['end_offset'] : null;
```
yguedidi: please keep this in the same line. BTW, you should cast to int the offset

src/Facebook/Exceptions/FacebookResumableUploadException.php
```diff
@@ -30,4 +30,39 @@
  */
 class FacebookResumableUploadException extends FacebookSDKException
 {
+    protected $startOffset;
+
+    protected $endOffset;
+
+    /**
+     * @return int
```
yguedidi: `int|null`

src/Facebook/Exceptions/FacebookResumableUploadException.php
```diff
@@ -30,4 +30,39 @@
...
+     */
+    public function getStartOffset()
+    {
+        return $this->startOffset;
+    }
+
+    /**
+     * @param int $startOffset
```
yguedidi: `int|null`

src/Facebook/Exceptions/FacebookResumableUploadException.php
```diff
@@ -30,4 +30,39 @@
...
+     */
+    public function setStartOffset($startOffset)
+    {
+        $this->startOffset = $startOffset;
+    }
+
+    /**
+     * @return int
```
yguedidi: `int|null`

src/Facebook/Exceptions/FacebookResumableUploadException.php
```diff
@@ -30,4 +30,39 @@
...
+     */
+    public function getEndOffset()
+    {
+        return $this->endOffset;
+    }
+
+    /**
+     * @param int $endOffset
```
yguedidi: `int|null`

src/Facebook/FileUpload/FacebookResumableUploader.php
```diff
@@ -121,6 +121,16 @@ public function transfer($endpoint, FacebookTransferChunk $chunk, $allowToThrow
                 throw $e;
             }
 
+            if (!is_null($preException->getStartOffset()) && !is_null($preException->getEndOffset())) {
```
yguedidi: I'd prefer `!== null` instead of `is_null`

tests/FileUpload/FacebookResumableUploaderTest.php
```diff
@@ -98,4 +98,14 @@ public function testFailedResumableTransferWillNotThrowAndReturnSameChunk()
...
     }
+
+    public function testFailedResumableTransferWillNotThrowAndReturnNewChunk()
+    {
+        $this->graphApi->failOnTransferAndUploadNewChunk();
+        $uploader = new FacebookResumableUploader($this->fbApp, $this->client, 'access_token', 'v2.4');
+
+        $chunk = new FacebookTransferChunk($this->file, '1', '2', '3', '4');
+        $newChunk = $uploader->transfer('/me/videos', $chunk);
+        $this->assertEquals('40', $newChunk->getStartOffset());
```
yguedidi: please assert also on the end ofset

yguedidi:
@phaizullin please rebase so that only your commit appear in the diff

yguedidi:
@phaizullin any news?

yguedidi:
Thank you @phaizullin

yguedidi:
And congrats for you first contribution to the SDK @phaizullin, I can't wait for your other pull requests!

PR #1008, opened 2018-06-14
alteracao da base

marciojsalmeida:
tirei  o www .


yguedidi:
@marciojsalmeida this is not a change we will accept. What the rational behind it? I'm closing this

PR #1018, opened 2018-06-27
target root directory for the better performance

voyula:


yguedidi:
@voyula Thanks! Have you got any link that shows the performance benefit of this change? :)

yguedidi:
Closed as no proof of performance gain

PR #1027, opened 2018-07-05
Update composer.json

ghost:
This update is fixing CPU usage and speed for autoloader.

Example; static and cached data.
Removes; checking namespaces - unnecessary continuous path parse - check class have or not for require - others...

In fact, **["optimize-autoloader"](https://getcomposer.org/doc/articles/autoloader-optimization.md)** is need all in the composer's.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,6 +34,9 @@
             "Facebook\\Tests\\": "tests/"
         }
     },
+    "config": {
+        "optimize-autoloader": true
+    },
     "extra": {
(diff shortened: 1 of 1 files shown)


ghost:
I'm joined in CLA.

ghost:
Hmm, i'm researched now again.

Not need if composer type is library because projects are can select this.
Only need use if composer type is project.

SammyK:
Thanks for the research! :)

PR #1028, opened 2018-07-06
Allow HHVM failures in TravisCI builds

SammyK:
Per [the discussion on the topic](https://github.com/facebook/php-graph-sdk/pull/854#issuecomment-401978940). We can always revert this later on if Facebook wants to jump in and maintain HHVM support. :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,8 @@ matrix:
   include:
     - php: hhvm
       dist: trusty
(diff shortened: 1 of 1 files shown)


yguedidi:
@SammyK I'd completly remove the HHVM job instead: having a test for something we know will always fail and we don't plan to fix is useless :)

PR #1034, opened 2018-07-25
Conserve id when next to data key, resolves #700

### LINKED ISSUE #700: Only "data" key is cast to GraphNode

### LINKED ISSUE #1010: facebook php sdk not returning page rating review id
When we calling the page ratings  fields "open_graph_story" from Facebook graph api then it is returning page ratings id but when we are calling from php sdk then it is coming as blank not returning  rating id so can not reply from api.
Please let me now if there is any other information I can provide. Thank you for your help.

jonny-no1:
closes #700, closes #1010

#700 is still consistently a problem across all versions of the Graph API.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Version 5 of the Facebook PHP SDK is a complete refactor of version 4. It comes 
   - Fixed HTTP/2 support (#1079)
   - Fixed resumable upload error (#1001)
   - Strip 'enforce_https' param (#1084)
+  - Conserve id when next to data key, resolves #700 (#1034)
 - 5.6.3 (2018-07-01)
   - Add fix for countable error in PHP 7.2 (originally #969 by @andreybolonin)
 - 5.6.2 (2018-02-15)
--- a/src/Facebook/GraphNodes/GraphNodeFactory.php
+++ b/src/Facebook/GraphNodes/GraphNodeFactory.php
@@ -304,7 +304,9 @@ class GraphNodeFactory
                 return $this->safelyMakeGraphEdge($data, $subclassName, $parentKey, $parentNodeId);
             }
             // Sometimes Graph is a weirdo and returns a GraphNode under the "data" key
-            $data = $data['data'];
+            $outerData = $data;
+            unset($outerData['data']);
+            $data = $data['data'] + $outerData;
         }
 
         // Create GraphNode
--- a/tests/GraphNodes/GraphNodeFactoryTest.php
+++ b/tests/GraphNodes/GraphNodeFactoryTest.php
@@ -247,6 +247,33 @@ class GraphNodeFactoryTest extends \PHPUnit_Framework_TestCase
         ], $graphData);
     }
 
+    public function testAGraphNodeWithARootDataKeyWillConserveRootKeys()
+    {
+        $data = json_encode([
+            'id' => '123',
+            'foo' => 'bar',
+            'data' => [
+                'name' => 'Foo McBar',
+                'link' => 'http://facebook/foo',
+            ],
+        ]);
+
+        $res = new FacebookResponse($this->request, $data);
+        $factory = new GraphNodeFactory($res);
+        $graphNode = $factory->makeGraphNode();
+
+        $this->assertInstanceOf('\Facebook\GraphNodes\GraphNode', $graphNode);
+
+        $graphData = $graphNode->asArray();
+
+        $this->assertEquals([
+            'id' => '123',
+            'foo' => 'bar',
+            'name' => 'Foo McBar',
+            'link' => 'http://facebook/foo',
+        ], $graphData);
+    }
+
     public function testAGraphEdgeWillBeCastRecursively()
     {
         $someUser = [


yguedidi:
@foaly-nr1 thanks for the PR! I'm closing it as #817 was opened before and that it's a more generic solution.
Your solution may lead to issue in case there is another `id` key under `data` key, so I prefer solution in #817

jonny-no1:
@yguedidi the documentation for `array_merge()` states:

> If the input arrays have the same string keys, then the later value for that key will overwrite the previous one.

So there wouldn't be an issue in case there is another `id` key under `data`.

Please consider reopening this PR. The other PR didn't pass the unit test and is now closed.

yguedidi:
@foaly-nr1 Yes I know, that's why it's an issue. Overwriting the `id` that is outside the `data` key is a loss of data. I answered in https://github.com/facebook/php-graph-sdk/pull/817#issuecomment-438944185 what should be the behavior IMO

yguedidi:
@foaly-nr1 here updated expected behavior
- if only a `data` key and no other keys (like `picture` in @tolbon above example): use the data for the node (`$data = $data['data']`)
- otherwise, use `data` key plus any other sibling key to `data`

So, something like:
```php
$outerData = $data;
unset($outerData['data']);
$data = $data['data'] + $outerData;
```
This keeps BC, and make most of key that were not accessible at all previous now accessible.
This doesn't solve the case where there is the same key next to `data` and inside `data`, but at least it's an improvement and should solve most issues.

Can you update you PR and target 5.x please? so this fix is part of the 5.7.0 release? Thanks

tests/GraphNode/GraphNodeFactoryTest.php
```diff
@@ -258,6 +258,30 @@ public function testAGraphNodeWithARootDataKeyWillBeCastAsAGraphNode()
         ], $graphData);
     }
 
+    public function testAGraphNodeWithARootDataKeyWillConserveId()
+    {
+        $data = json_encode([
+            'id' => '123',
```
yguedidi: please add another key next do `id` and `data` here

```suggestion
            'id' => '123',
            'foo' => 'bar',
```

tests/GraphNode/GraphNodeFactoryTest.php
```diff
@@ -258,6 +258,30 @@ public function testAGraphNodeWithARootDataKeyWillBeCastAsAGraphNode()
...
+            ],
+        ]);
+
+        $res = new Response($this->request, $data);
+
+        $factory = new GraphNodeFactory($res);
+        $graphNode = $factory->makeGraphNode();
+        $graphData = $graphNode->asArray();
+
+        $this->assertInstanceOf(GraphNode::class, $graphNode);
```
yguedidi: please don't use `::class`, SDK v5 supports PHP 5.4

tests/GraphNode/GraphNodeFactoryTest.php
```diff
@@ -258,6 +258,30 @@ public function testAGraphNodeWithARootDataKeyWillBeCastAsAGraphNode()
         ], $graphData);
     }
 
+    public function testAGraphNodeWithARootDataKeyWillConserveId()
```
yguedidi:
```suggestion
    public function testAGraphNodeWithARootDataKeyWillConserveRootKeys()
```

yguedidi (changes requested):
Please rebase on `5.x` branch, target it in the PR and update the changelog too (use PR number in the changelog)

src/GraphNode/GraphNodeFactory.php
```diff
@@ -301,7 +301,7 @@ public function castAsGraphNodeOrGraphEdge(array $data, $subclassName = null, $p
                 return $this->safelyMakeGraphEdge($data, $subclassName, $parentKey, $parentNodeId);
             }
             // Sometimes Graph is a weirdo and returns a GraphNode under the "data" key
-            $data = $data['data'];
+            $data = array_merge((isset($data['id']) ? ['id' => $data['id']] : []), $data['data']);
```
yguedidi:
```suggestion
            $outerData = $data;
            unset($outerData['data']);
            $data = $data['data'] + $outerData;
```

tests/GraphNode/GraphNodeFactoryTest.php
```diff
@@ -258,6 +258,30 @@ public function testAGraphNodeWithARootDataKeyWillBeCastAsAGraphNode()
...
+
+        $res = new Response($this->request, $data);
+
+        $factory = new GraphNodeFactory($res);
+        $graphNode = $factory->makeGraphNode();
+        $graphData = $graphNode->asArray();
+
+        $this->assertInstanceOf(GraphNode::class, $graphNode);
+        $this->assertEquals([
+            'id' => '123',
```
yguedidi:
```suggestion
            'id' => '123',
            'foo' => 'bar',
```

jonny-no1:
All done and change log updated.

yguedidi (approved):

jonny-no1:
It's been merged!! Fittingly it was the 2 year anniversary of my bug report of this issue last week.

PR #1039, opened 2018-08-09
Update GraphPage::getPerms return type documentation

signe:
Update GraphPage::getPerms return type documentation


yguedidi:
Thanks @signe! Can you point me to a place in the official doc that shows that? :)

PR #1145, opened 2019-08-04
[6.0] Use PSR 17 and 18 standard. 

Girgias:
Also small improvements in Facebook\Client to handle PSR-7 messages.

This also changes multiple constructors for Dependency Injection.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,16 +16,15 @@
     },
     "require": {
         "php": "^7.1",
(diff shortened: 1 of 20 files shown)


Girgias:
I'm thinking of removing the setter and getter method for the HTTP Client in ``Facebook\Client`` as it is injected and end users would already know/have a PSR-18 HTTP Client

PR #1251, opened 2021-09-06
Added PHP 8.0 to composer.json

tnhnclskn:


ryanwcare2 (approved):
No one has chimed in yet, but this patch works perfectly for me with PHP 8. Can someone merge this? It would help a lot of people out.

JexPY:
@ryanwcare2 
> No one has chimed in yet, but this patch works perfectly for me with PHP 8. Can someone merge this? It would help a lot of people out.

As this project is moved to the archive I don't think that anyone will merge the contributions, so I moved to [https://github.com/facebook/facebook-php-business-sdk](https://github.com/facebook/facebook-php-business-sdk) maybe it will suit you too.

I think we should leave this project as we see Facebook killed it and we are trying to wake the dead one up.

PR #1257, opened 2021-10-13
Added CURL option to force TLS version 1.2

vladget:
For securing platform data we added CURL option to force TLS version 1.2 in FacebookCurlHttpClient


vladget:
Good day.
We signed CLA. Please check out our pull request.

